PR #47, opened 2022-05-25
Fix the path to health endpoint of backend

emanueldfine:
# Pull Request \It's impossible that the actual e2etests start because the script which checks all services won't reach the backend health endpoint
`<Description here>`
- [ ] A peer-review has been executed
  - [ ] The new feature was observed "in running software"
- [ ] Documentation is updated as required
- [ ] The automated deployment is updated if required
- [ ] The new version is deployed to the dev server using this branch
  - [ ] It's verified that this version actually is the one deployed (check actuator/info for branch name and commit id!)
  - [ ] The new feature is manually used/tested/observed on dev server
- [ ] There is at least one picture for each story, which was created before coding has started
- [ ] The local Dev stack still works: execute `startDevelopmentStack.sh`, npm serve, npm teste2e and execute Cypress Tests locally

### DIFF
--- a/testing/wait_for_backend_and_run_e2etests.sh
+++ b/testing/wait_for_backend_and_run_e2etests.sh
@@ -6,7 +6,7 @@ set -ex
 
 is_infrastructure_up () {
   declare -A services
-  services["backend"]=http://proxy:80/api/actuator/health/ping
+  services["backend"]=http://proxy:80/api/actuator/health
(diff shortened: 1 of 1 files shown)


emanueldfine:
Closing PR because it is in fact better to access `/actuator/health/ping` than directly `/actuator/health`.
Reason: `actuator/health` has multiple occurrences of the string "UP". Therefore it is not assured that we check for the right "UP" when we `grep` the response of this endpoint.

PR #105, opened 2022-09-28
Dala 555 make keycloak settings updatable

ppuetsch:
# Pull Request \<Title>
`<Description here>`
- [x] The new version is deployed to the dev server using this branch
  - [x] It's verified that this version actually is the one deployed (check actuator/info for branch name and commit id!)
- [ ] There is at least one picture for each story, which was created before coding has started
- [x] The local Dev stack still works: execute `startDevelopmentStack.sh`, execute npm run cypress and run Cypress Tests locally
- [ ] After merge, remove github secret RESET_KEYCLOAK_SECURITY_PHRASE_SECRET


dataland-keycloak/start_keycloak.sh
```diff
@@ -1,14 +1,29 @@
...
-if [[ "$1" == initialize ]]; then
+if [[ "$mode" == initialize ]]; then
   echo "Initializing new keycloak realms"
- ./kc.sh import --file /opt/keycloak/data/import/master-realm.json
- ./kc.sh start --import-realm
+  mkdir -p $dataland_realm_folder
+  cp /keycloak_users/datalandsecurity-users-*.json $dataland_realm_folder || echo "No importable users exist"
+  rm $(grep -E -l '"username" : "data_(reader|uploader)"' $dataland_realm_folder/datalandsecurity-users-*.json) || echo "No technical users to be cleaned up"
+  rm $(grep -E -l '"username" : "test_user.*@dataland.com"' $dataland_realm_folder/datalandsecurity-users-*.json) || echo "No test users to be cleaned up"
+  cp /keycloak_realms/datalandsecurity-realm.json $dataland_realm_folder
```
F2-6Reck, line 15: Is this old information anyway overwritten with the restart or why is it not removed?

dataland-keycloak/start_keycloak.sh
```diff
@@ -1,14 +1,29 @@
...
+  for variable in $(env | grep KEYCLOAK_ | cut -d'=' -f1); do
+    #If one of the env variables containing "KEYCLOAK_" contains a % character the below sed statement will fail and the script will terminate
```
F2-6Reck, line 17: Why is this important?

deployment/initialize_keycloak.sh
```diff
@@ -39,6 +50,19 @@ timeout 300 bash -c "while ! ssh ubuntu@\"$target_server_url\" \"cd $location &&
                        sleep 5;
                      done"
 
+if ssh ubuntu@"$target_server_url" "ls $keycloak_user_dir/*-users-*.json &>/dev/null"; then
+  echo "Testing if the number of current users matches the number of exported users"
+  current_users=$(ssh ubuntu@"$target_server_url" "docker exec $container_name /opt/keycloak/bin/kcadm.sh get users -r datalandsecurity --server http://localhost:8080/keycloak --realm master --user $KEYCLOAK_ADMIN --password $KEYCLOAK_ADMIN_PASSWORD | grep -c '\"username\" :'")
```
F2-6Reck, line 55: What are the current users / how are they different from all users?
If current users = all users \ test users, why do we need the all users as separate entitity?

dataland-keycloak/start_keycloak.sh
```diff
@@ -1,14 +1,29 @@
 #!/bin/bash
 set -eu
 
+mode=$1
+dataland_realm_folder=/opt/keycloak/datalandsecurity
```
F2-6Reck, line 5: So, in the end everything is copied here and when something goes wrong, this could serve as a backup? And this will be overwritten whenever keycloak is restarted?
Should there be something that makes checkpoints that can serve as further backups? Security issues?

dataland-keycloak/start_keycloak.sh
fsoffel: The environment variables need to be inserted manually since keycloak only does not do it on its own using the import command.

deployment/initialize_keycloak.sh
fsoffel: This is actually a test that can only be executed at this point of the keycloak initialization. We are comparing the number of users that are currently stored in the keycloak docker volume (current_users) vs the number of users that should have been imported (expected_users=all_users-test_users) where test_users is the number of users generated by E2E tests and therefore should and are not imported again.

dataland-keycloak/start_keycloak.sh
fsoffel: The backup folder is filled and cleared (mv) in subsequent commands. So there hopefully should not be too much of a security risk. This backup folder is only relevant if initialize_keycloak.sh fails at a point where the keycloak docker volume does not exist, i.e. after the volume gets deleted and before the time out loop waiting for the keycloak-initializer to finish initialization terminates without time out. In this case the /home/ubuntu/dataland/dataland-keycloak/users persists until the next execution of start_and_deploy_to_server.sh where it gets deleted and recreated empty. That is where we use the backup folder.

dataland-keycloak/start_keycloak.sh
fsoffel: The only container having access to this data keycloak-initializer which gets shut down in initialize_keycloak.sh. The corresponding folder on the server is cleaned up in start_and_deploy_to_server.sh

dataland-keycloak/start_keycloak.sh
```diff
@@ -1,14 +1,29 @@
...
 cd $script_dir
 
-if [[ "$1" == initialize ]]; then
+if [[ "$mode" == initialize ]]; then
   echo "Initializing new keycloak realms"
- ./kc.sh import --file /opt/keycloak/data/import/master-realm.json
- ./kc.sh start --import-realm
+  mkdir -p $dataland_realm_folder
+  cp /keycloak_users/datalandsecurity-users-*.json $dataland_realm_folder || echo "No importable users exist"
+  rm $(grep -E -l '"username" : "data_(reader|uploader)"' $dataland_realm_folder/datalandsecurity-users-*.json) || echo "No technical users to be cleaned up"
```
F2-6Reck, line 13: Does this container need to be cleared at some place?

F2-6Reck: Seems to be happening automatically.

dataland-keycloak/start_keycloak.sh
SiegfriedSobkowiak: the sed command needs a separator (that can be any character), however the separation character cannot be in the search and replace values since the command crashes otherwise. The usual separator is / however the frontend url does contain / and hence we decided to go with % instead. The comment is intended for other devs to point out that this might be an issue since debugging this would be cumbersome at best.

deployment/deployment_utils.sh
```diff
@@ -11,14 +11,53 @@ wait_for_health () {
...
 
+kill_docker_containers() {
+  docker kill $(docker ps -q); docker system prune --force; docker info
+}
+
 delete_docker_volume_if_existent () {
-  target_server_url=$1
-  location=$2
-  volume_filter=$3
+  volume_filter=$1
```
ppuetsch, line 19: Das ist ja ein Util - das Util sollte nciht davon ausgehen, dass irgendein Parameter tatsächlich an einer Position als Command Line Parameter steht. 

Ich fände besser, wenn volume_filter ein parameter von delete_docker_volume_if_existent wäre, und wenn die aufrufende Stelle dann $1 hier rein wirft.

deployment/deployment_utils.sh
```diff
@@ -11,14 +11,53 @@ wait_for_health () {
...
+  old_volume=$(search_volume "$volume_filter")
   if [[ -n $old_volume ]]; then
     echo "Removing old database volume with name $old_volume."
-    ssh ubuntu@"$target_server_url" "cd $location && sudo docker volume rm $old_volume"
+    docker volume rm "$old_volume"
   fi
 }
+
+search_volume() {
+  volume_filter=$1
```
ppuetsch, line 29: wie oben - volume_filter als Funktionsparameter

ppuetsch: Das war quatsch was ich geschrieben habe. Bitte ignorieren.

deployment/deployment_utils.sh
ppuetsch: Das war quatsch was ich geschrieben habe. Bitte ignorieren.

deployment/deployment_utils.sh
```diff
@@ -11,14 +11,53 @@ wait_for_health () {
   timeout 240 bash -c "while ! curl -L $1 2>/dev/null | grep -q UP; do echo 'Waiting for $2 to finish boot process.'; sleep 5; done; echo '$2 available!'"
 }
 
+kill_docker_containers() {
+  docker kill $(docker ps -q); docker system prune --force; docker info
+}
+
 delete_docker_volume_if_existent () {
```
ppuetsch, line 18: warum nicht einfach 
```
docker volume rm $1 || true
```

fsoffel: Lesbarkeit im Zusammenspiel mit dem echo command.

PR #194, opened 2023-03-15
Dala 1318 fix e2e tests

ahoecherl:
# DALA-1318, fix e2e tests
Making e2e tests for stable. General issue is that dummy QA processing may not yet have been completed when data is accessed. In e2e testing, this can be solved by shorting briefly after an upload.
Adding a wait blanket wait statement after upload increased the overall runtime of the e2e API tests by about 5 seconds.


dataland-backend/src/main/kotlin/org/dataland/datalandbackend/services/DataManager.kt
```diff
@@ -241,7 +251,7 @@ class DataManager(
     fun getDataSet(dataId: String, dataType: DataType, correlationId: String): StorableDataSet {
         assertActualAndExpectedDataTypeForIdMatch(dataId, dataType, correlationId)
         val dataMetaInformation = metaDataManager.getDataMetaInformationByDataId(dataId)
-        val dataAsString = getDataFromStorage(dataId, correlationId)
+        val dataAsString = getDataFromStorageLocation(dataId, correlationId)
```
ppuetsch, line 254: Can we improve naming?
like: 
`getDataFromStorage` -> `getDataFromStorageService`
`getDataFromStorageLocation` -> `getDataFromCacheOrStorageService`

I am happy for better suggestions - however `getDataFromStorageLocation` doesn't make me expect what the method actually does.

dataland-backend/src/test/kotlin/org/dataland/datalandbackend/services/DataManagerTest.kt
```diff
@@ -106,6 +99,7 @@ class DataManagerTest(
             storableEuTaxonomyDataSetForNonFinancials, correlationId,
         )
         `when`(mockStorageClient.selectDataById(dataId, correlationId)).thenReturn("")
+        dataManager.removeStoredItemFromTemporaryStore(dataId, "", MessageType.DataStored)
```
ppuetsch, line 102: Isn't that redundant? BeforeEach the dataManager will be reset, and then it's again guaranteed that the dataId is removed.

dataland-e2etests/src/test/kotlin/org/dataland/e2etests/utils/ApiAccessor.kt
```diff
@@ -122,6 +126,8 @@ class ApiAccessor {
                 )
             }
         }
+//        Sleep after upload to ensure that dummy QA is passed and data actually available for further processing
+        if (sleepAfterUpload) Thread.sleep(sleepAfterUploadMs)
```
ppuetsch, line 130: I don't like random sleeps. As this is a test, I don't care too much. But I would prefer if we could introduce some waiting mechanism that waits for the right thing to be true (e.g. that a dataset is QAed), and not just waiting a fixed amount of time. 

What do you think?

dataland-internal-storage/src/main/kotlin/org/dataland/datalandinternalstorage/services/DatabaseDataStore.kt
```diff
@@ -83,12 +85,22 @@ class DatabaseDataStore(
...
+    }
+
     /**
      * Reads data from a database
      * @param dataId the id of the data to be retrieved
      * @return the data as json string with id dataId
      */
     fun selectDataSet(dataId: String): String {
-        return dataItemRepository.findById(dataId).orElse(DataItem("", "")).data
+        return dataItemRepository.findById(dataId).orElse(DataItem("", "\"\"")).data
```
ppuetsch, line 104: Was that even right behaviour in the first place? Shouldn't we return a 404 in the case a data Id is not present - not not simply return an empty object?

fsoffel: I would also prefer that.

dataland-backend/src/main/kotlin/org/dataland/datalandbackend/services/DataManager.kt
fsoffel: 👍

dataland-backend/src/test/kotlin/org/dataland/datalandbackend/services/DataManagerTest.kt
fsoffel: This is to prevent running into the case that the data is still in the map and the StorageControllerApi is actually used.

dataland-backend/src/main/kotlin/org/dataland/datalandbackend/services/DataManager.kt
fsoffel: Done

dataland-internal-storage/src/main/kotlin/org/dataland/datalandinternalstorage/services/DatabaseDataStore.kt
fsoffel: Done

ppuetsch: As agreed: 
* don't use 404, but HttpStatus.NOT_FOUND
* Throw a ResourceNotFoundException if the optional from database is not present

dataland-backend/src/main/kotlin/org/dataland/datalandbackend/services/DataManager.kt
ppuetsch: ok

dataland-backend/src/test/kotlin/org/dataland/datalandbackend/services/DataManagerTest.kt
ppuetsch: ok

dataland-internal-storage/src/main/kotlin/org/dataland/datalandinternalstorage/services/DatabaseDataStore.kt
ppuetsch: ok

dataland-e2etests/src/test/kotlin/org/dataland/e2etests/utils/ApiAccessor.kt
ahoecherl: Function has been added that explicitly checks for each upload item if qa is set to passed and waits and retries until this is the case for all uploaded items.

dataland-e2etests/src/test/kotlin/org/dataland/e2etests/utils/ApiAccessor.kt
```diff
@@ -122,7 +124,32 @@ class ApiAccessor {
                 )
             }
         }
-        return listOfUploadInfo
+        if (!ensureQaPassed) return listOfUploadInfo
+        return ensureQaCompleted(listOfUploadInfo)
+    }
+
+    private fun ensureQaCompleted(listOfUploadInfo: MutableList<UploadInfo>): MutableList<UploadInfo> {
+        // Check after upload to ensure that dummy QA is passed and wait until it is
```
ppuetsch, line 132: Please use docstring instead of a comment in line 1 of a method

dataland-e2etests/src/test/kotlin/org/dataland/e2etests/utils/ApiAccessor.kt
```diff
@@ -122,7 +124,32 @@ class ApiAccessor {
...
+        val maxQaPassedYetRetries = 100; val sleepIfQaNotPassedYetMs: Long = 50
+
+        val result = mutableListOf<UploadInfo>()
+        repeat(maxQaPassedYetRetries) {
```
ppuetsch, line 136: Isn't awaitility what you want here?
https://github.com/awaitility/awaitility/wiki/Usage#simple

dataland-e2etests/src/test/kotlin/org/dataland/e2etests/utils/ApiAccessor.kt
```diff
@@ -122,7 +124,32 @@ class ApiAccessor {
...
+                    listOfUploadInfo.removeAt(idx)
+                }
+            }
+            if (listOfUploadInfo.isEmpty()) return result
+            Thread.sleep(sleepIfQaNotPassedYetMs)
+        }
+        class QaNotCompletedException(message: String) : RuntimeException(message)
+        throw QaNotCompletedException(
+            "$maxQaPassedYetRetries retries every $sleepIfQaNotPassedYetMs ms, qa for $listOfUploadInfo still failed",
+        )
```
ppuetsch, line 152: What about this approach: 
A list of all dataIds is collected. Now it's checked for every ID if QA is passed. If not, this ID is put to a new list. 
In the End, the new list is assigned to the old list and list.length() == 0 is returned. 

this method now goes into Awaitility

ahoecherl: Was refactored in the meantime due to Sonarcube complains. 

Operating on list of dataIds is difficult since UploadInfo has a lot more than just the dataIds and keeping track between the mapping of each dataId to the associated UploadInfo would require quite some effort. Therefore I've instead changed the metaData in UploadInfo to be a var and operate on UploadInfo objects directly.

Other than this the done refactoring into method updateQaStatus seems similar to what you propose.

dataland-e2etests/src/test/kotlin/org/dataland/e2etests/utils/ApiAccessor.kt
ahoecherl: I would want to also achieve that the UploadInfo objects are updated with their current Metadata. If I just run await until a method returns true for a list of dataIds, then the UploadInfo Objects still contain Metadata that states that the QA is Pending even though it is not.

dataland-e2etests/src/test/kotlin/org/dataland/e2etests/utils/ApiAccessor.kt
```diff
@@ -122,7 +124,32 @@ class ApiAccessor {
...
+        val result = mutableListOf<UploadInfo>()
+        repeat(maxQaPassedYetRetries) {
+            for (idx in listOfUploadInfo.indices.reversed()) {
+                var metaData = listOfUploadInfo[idx].actualStoredDataMetaInfo!!
+                listOfUploadInfo[idx].actualStoredDataMetaInfo = metaDataControllerApi.getDataMetaInfo(metaData.dataId)
+                metaData = listOfUploadInfo[idx].actualStoredDataMetaInfo!!
+                if (metaData.qaStatus == QAStatus.accepted) {
+                    result.add(listOfUploadInfo[idx])
+                    listOfUploadInfo.removeAt(idx)
+                }
```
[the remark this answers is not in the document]
ahoecherl: the .reverse solves this issue but has been refactored away nevertheless.

dataland-e2etests/src/test/kotlin/org/dataland/e2etests/utils/ApiAccessor.kt
ahoecherl: Has been refactored to use await

dataland-e2etests/src/test/kotlin/org/dataland/e2etests/utils/ApiAccessor.kt
ahoecherl: Has been refactored as proposed

dataland-e2etests/src/test/kotlin/org/dataland/e2etests/utils/ApiAccessor.kt
ahoecherl: using await now.

settings.gradle.kts
```diff
@@ -53,6 +53,7 @@ dependencyResolutionManagement {
 
             library("bcpkix-jdk15on", "org.bouncycastle:bcpkix-jdk15on:1.70")
             library("bcprov-jdk15on", "org.bouncycastle:bcprov-jdk15on:1.70")
+            library("awaitility", "org.awaitility:awaitility:4.2.0")
```
ppuetsch, line 56: As you anyways don't use `libs.awaitility`, this row is unneccesarry.

dataland-e2etests/build.gradle.kts
```diff
@@ -31,6 +31,7 @@ dependencies {
     implementation(libs.log4j.to.slf4j)
     implementation("org.springframework.boot:spring-boot-starter-web")
     testImplementation("org.springframework.boot:spring-boot-starter-test")
+    testImplementation("org.awaitility:awaitility")
```
ppuetsch, line 34: here you rely on the version specified by Spring (since you don't provide a version explicitly). Hence the entry in settings.gradle not used, so it can be removed there.

dataland-e2etests/src/test/kotlin/org/dataland/e2etests/utils/ApiAccessor.kt
```diff
@@ -122,9 +127,38 @@ class ApiAccessor {
...
+                )
+            if (metaData.qaStatus != QAStatus.accepted) {
+                uploadInfo.actualStoredDataMetaInfo = metaDataControllerApi.getDataMetaInfo(metaData.dataId)
+                if (uploadInfo.actualStoredDataMetaInfo!!.qaStatus != QAStatus.accepted) {
+                    allQaPassedIndicator = false
+                }
+            }
+        }
+        return allQaPassedIndicator
+    }
```
ppuetsch, line 160: The Pattern "set a boolean to an initial value, and reverse it's value on a specific condition afterwards, and that's what you return" can be avoided (as it's not so nice) by using the streams-equivalent of Kotlin.

So this method can be simplified to:
```
    private fun checkIfQaPassedAndUpdateMetadata(uploadInfos: List<UploadInfo>): Boolean {
        return uploadInfos.all { uploadInfo ->
            val metaData = uploadInfo.actualStoredDataMetaInfo
                    ?: throw NullPointerException(
                            "To check QA Status, metadata is required but was null for $uploadInfo",
                    )
            if (metaData.qaStatus != QAStatus.accepted) {
                uploadInfo.actualStoredDataMetaInfo = metaDataControllerApi.getDataMetaInfo(metaData.dataId)
            }
            return uploadInfo.actualStoredDataMetaInfo!!.qaStatus == QAStatus.accepted
        }
    }
```

Probably it's even a little bit more ressource efficient as it stops processing once the first updated QA Status is != Accepted.

ahoecherl: since in Kotlin streams on Collections are eager instead of lazy my argument was wrong. I've adapted the proposed stream-based implementation.

dataland-e2etests/src/test/kotlin/org/dataland/e2etests/utils/ApiAccessor.kt
ppuetsch: ok

dataland-e2etests/src/test/kotlin/org/dataland/e2etests/utils/ApiAccessor.kt
ppuetsch: ok

dataland-e2etests/src/test/kotlin/org/dataland/e2etests/utils/ApiAccessor.kt
ppuetsch: ok

dataland-e2etests/src/test/kotlin/org/dataland/e2etests/utils/ApiAccessor.kt
ppuetsch: ok

settings.gradle.kts
ppuetsch: fixed

dataland-e2etests/build.gradle.kts
ppuetsch: done

dataland-e2etests/src/test/kotlin/org/dataland/e2etests/utils/ApiAccessor.kt
ppuetsch: ok

PR #197, opened 2023-03-22
Dala 1344 adjust rabbit mq disk and ram usage

ahoecherl:
# Pull Request \<Title>
`<Description here>`

### DIFF
--- a/dataland-rabbitmq/rabbitmq.conf
+++ b/dataland-rabbitmq/rabbitmq.conf
@@ -1,2 +1,3 @@
 management.load_definitions = /etc/rabbitmq/definitions.json
+# Always maintain 1.5 ratio of disk_free_limit.absolute to RAM allocated to rabbitMQ container.
 disk_free_limit.absolute = 768MB
--- a/dataland-rabbitmq/rabbitmq.conf
+++ b/dataland-rabbitmq/rabbitmq.conf
@@ -1,2 +1,2 @@
 management.load_definitions = /etc/rabbitmq/definitions.json
-disk_free_limit.relative=1.5
+disk_free_limit.absolute = 768MB
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -289,6 +289,7 @@ services:
     deploy:
       resources:
         limits:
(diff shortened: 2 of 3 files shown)


dataland-rabbitmq/rabbitmq.conf
```diff
@@ -1,2 +1,2 @@
 management.load_definitions = /etc/rabbitmq/definitions.json
-disk_free_limit.relative=1.5
+disk_free_limit.absolute = 768MB
```
SiegfriedSobkowiak, line 2: From the Jira-Ticket: "512 MB RAM and 1,5*512MB for disk space and add a comment to change either one if the other one is changed" I do not see a comment like that anywhere

docker-compose.yml
```diff
@@ -286,6 +286,10 @@ services:
       test: rabbitmq-diagnostics -q alarms
       interval: 30s
       timeout: 3s
+    deploy:
+      resources:
+        limits:
+          memory: 512M
```
SiegfriedSobkowiak, line 292: Maybe I am just confused because I was not part of the conversation, but: Why is the max memory usage of the container times 1,5 the minimum required free disk space? Is that to be able to dump the memory to disk in case of a certain event?

ppuetsch: I think because the rabbitmq-prepare-for-production guideline tells to do so. 
Reasoning probably is: With X RAM RabbitMQ can process messages up to amount X. As rabbitMQ guarantees persistence, it therefore needs X+overhead disk space.

ppuetsch:
image

Seems to work. :-)

docker-compose.yml
ahoecherl: Exactly as @ppuetsch said. This is the reason: https://www.rabbitmq.com/production-checklist.html#resource-limits-disk-space

dataland-rabbitmq/rabbitmq.conf
ahoecherl: Sorry. Comments have been added and pushed now.

PR #222, opened 2023-05-12
Dala 1326 upload real data pdfs

SiegfriedSobkowiak:
# Pull Request \<Title>
`<Description here>`
- [x] The new version is deployed with fake data to the dev server "dev1" using this branch
  - [x] Run with setting `Use real data` turned off and `Reset backend database & Re-populate` turned on 
- [x] The new version is deployed with real data to a dev server using this branch
  - [x] Run with both settings `Reset backend database & Re-populate` and `Use real data` turned on


.github/workflows/CD.yaml
```diff
@@ -171,7 +171,8 @@ jobs:
         run: |
           if [[ $REALDATA == true ]]; then
             echo "Retrieving real data"
-            docker run -v $(pwd)/testing/data:/testData:rw ghcr.io/d-fine/dataland/datacontainer:0.0.17
+            rm -rf ./testing/data/documents/*
```
stewal19, line 174: In principle everytime a deployment is done all existing documents are deleted and then the new ones will be uploaded, correct?

dataland-csvconverter/src/main/kotlin/org/dataland/csvconverter/csv/commonfieldparsers/EuTaxonomyCommonFieldParser.kt
```diff
@@ -49,7 +49,7 @@ class EuTaxonomyCommonFieldParser(
     /**
      * Returns the number of employees for a specific company
      */
-    fun getNumberOfEmployees(csvLineData: Map<String, String>): BigDecimal {
-        return columnMappingEuTaxonomyUtils.readCsvDecimal("numberOfEmployees", csvLineData)!!
+    fun getNumberOfEmployees(csvLineData: Map<String, String>): BigDecimal? {
```
stewal19, line 52: Is this due to some companies not having information about employees?

dataland-csvconverter/src/main/kotlin/org/dataland/csvconverter/csv/commonfieldparsers/FiscalYearParser.kt
```diff
@@ -27,16 +27,16 @@ class FiscalYearParser {
     /**
      * Method to retrieve information about the deviation of the fiscal year from the csv file
      */
-    fun getFiscalYearDeviation(csvLineData: Map<String, String>): FiscalYearDeviation {
-        return columnMappingFiscalYear.getCsvValue("fiscalYearDeviation", csvLineData)
-            .let { fiscalYearDeviationParser.parse("fiscalYearDeviation", it) }
+    fun getFiscalYearDeviation(csvLineData: Map<String, String>): FiscalYearDeviation? {
```
stewal19, line 30: Is this due to not availaible information about fiscal year deviation / fiscal year end?

dataland-csvconverter/pdfMapping.sh
```diff
@@ -0,0 +1,35 @@
+#!/bin/bash
```
stewal19, line 1: The connection between the reports and the dataset is yet to be made, right? 
So after running this script we have the mapping between the dataset and the connected reports but still have to add this information to the dataset. Or have i missed something?

.github/workflows/CD.yaml
SiegfriedSobkowiak: Yes that is the idea. Reuse the existing mechanism and simply overwrite the deployed data with the "real data" content.

dataland-csvconverter/src/main/kotlin/org/dataland/csvconverter/csv/commonfieldparsers/EuTaxonomyCommonFieldParser.kt
SiegfriedSobkowiak: The issue here was that fields were flagged as required in the data model but we do not have real data for them. Hence we reverted the change in the data model but nobody reverted the change in the csv converter. Which I now did in this PR so we are able to parse CSVs again.

dataland-csvconverter/pdfMapping.sh
SiegfriedSobkowiak: I will document the process in the wiki once I get the time. The idea is the following: running this script collects the pdfs in the folder and it creates the output csv file. The content of that file is copy-pasted into the data dictionary. From there the csv is exported again as input for the other already existing bash script that will create the data for the container and push it. No need to do anything with the data sets.

SiegfriedSobkowiak: Maybe we should keep this open as a reminder until the documentation is updated.

dataland-csvconverter/src/main/kotlin/org/dataland/csvconverter/csv/commonfieldparsers/EuTaxonomyCommonFieldParser.kt
stewal19: Ok

.github/workflows/CD.yaml
stewal19: Thanks for clarification

dataland-csvconverter/pdfMapping.sh
SiegfriedSobkowiak: I updated the documentation in section "Real-Data-Container". Can you have a look and tell me if it is clear?

dataland-frontend/tests/e2e/specs/index.ts
```diff
@@ -24,6 +24,8 @@ require("./infrastructure");
 if (runPrepopulation) {
   if (!singlePopulate || cypressTestGroup === 1 || cypressTestGroup === 101) {
     require("./prepopulation");
+  } else if (cypressTestGroup === 3) {
```
stewal19, line 27: Is this only for logging or does test group 3 now run before prepopulation is finished?

dataland-frontend/cypress.config.ts
```diff
@@ -121,6 +121,12 @@ export default defineConfig({
                     return createHash("sha256").update(file).digest("hex");
                 },
             });
+            on("task", {
```
stewal19, line 124: What arguments are looged here?

dataland-csvconverter/pdfMapping.sh
```diff
@@ -0,0 +1,48 @@
...
+
+output_folder=$(dirname "$0")/pdfs
+mkdir -p "$output_folder"
+rm "$output_folder"/*.pdf || true
+output_file=$output_folder/output.csv
+
+echo "Annual Report File,Sustainability Report File,Integrated Report File,Annual Report,Sustainability Report,Integrated Report" > "$output_file"
+sed 1d "$csv_file" | while read -r line; do
+  isin=$(echo "$line" | grep -oP ';([A-Z0-9]{12});' | cut -d';' -f2) || true
+  if [[ -z "$isin" ]]; then
```
stewal19, line 26: If i understood it correctly, if a company does not have an ISIN this script does not work, as there is no other way to connect the report to the company entry. So in the future this might has to change when other identifiers are also used. Maybe as a loop through all existing identifiers of a company and check for every ID if there is a report corresponding to it.

dataland-csvconverter/pdfMapping.sh
stewal19: Ok

dataland-frontend/cypress.config.ts
stewal19: Discussed in a call - ok

PR #229, opened 2023-05-16
DALA-1213 Further UI improvements

MarcTM01:
# Pull Request \<Title>
`<Description here>`
- [x] The new version is deployed with fake data to the dev server "dev1" using this branch
  - [x] Run with setting `Use real data` turned off and `Reset backend database & Re-populate` turned on 
- [x] The new version is deployed with real data to a dev server using this branch
  - [x] Run with both settings `Reset backend database & Re-populate` and `Use real data` turned on


dataland-backend/src/main/kotlin/org/dataland/datalandbackend/model/lksg/categories/general/subcategories/LksgProductionSpecific.kt
```diff
@@ -4,7 +4,6 @@ import com.fasterxml.jackson.annotation.JsonProperty
 import org.dataland.datalandbackend.model.enums.commons.YesNo
 import org.dataland.datalandbackend.model.enums.lksg.NationalOrInternationalMarket
 import org.dataland.datalandbackend.model.lksg.LksgProductionSite
-import java.math.BigDecimal
```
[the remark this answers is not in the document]
emanueldfine: (Together with `LksgProductionSpecific`)

dataland-frontend/src/components/resources/frameworkDataSearch/lksg/LksgDataModel.ts
```diff
@@ -28,54 +28,48 @@ export const lksgDataModel = [
               "Is your head office, administrative headquarters, registered office or subsidiary located in Germany?",
             label: "Head Office in Germany",
             component: "YesNoFormField",
-            required: true,
-            validation: "required",
+            required: false,
           },
           {
             showIf: (): boolean => true,
```
emanueldfine, line 34: In the CreateLksgData.vue component, you do
`v-if="field.showIf(companyAssociatedLksgData.data)"`
The IDE is complaining that `showIf()` does not expect an input argument, which is correct for all subcategories which have `showIf: (): boolean => true`.

Shouldn't we make `showIf` accept an input everywhere, and hardcode a return value `true` for those subcategories, which should always be visible?

dataland-frontend/src/components/resources/frameworkDataSearch/lksg/LksgDataModel.ts
```diff
@@ -28,54 +28,48 @@ export const lksgDataModel = [
...
             name: "groupOfCompanies",
             description: "Do you belong to a group of companies?",
             label: "Group of Companies",
             component: "YesNoFormField",
-            required: true,
-            validation: "required",
+            required: false,
           },
           {
             showIf: (dataModel: LksgData): boolean => dataModel?.general?.masterData?.groupOfCompanies === "Yes",
```
emanueldfine, line 42: Furthermore, I don't really like the name `showIf`.
It really confused me when I read it for the first time.
Some alternative name suggestions:

`isSubcategoryDisplayedInForm()`
`isDisplayedInForm()`
`isDisplayed()`
`isSubcategoryRendered()`
...

dataland-frontend/tests/e2e/fixtures/lksg/LksgDataFixtures.ts
```diff
@@ -116,7 +116,7 @@ export function generateLksgData(undefinedProbability = 0.5): LksgData {
       },
       productionSpecific: {
         manufacturingCompany: valueOrUndefined(randomYesNo(), undefinedProbability),
-        capacity: valueOrUndefined(randomNumber(100000), undefinedProbability),
+        capacity: valueOrUndefined(randomNumber(100000).toString(), undefinedProbability),
```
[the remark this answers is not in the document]
janherre: Isn't this an actual string already? Should we implement a faker that can produce a number with unit as a string? Because just using a string faker wont create any more realistic fake fixtures, I think.

dataland-backend/src/main/kotlin/org/dataland/datalandbackend/model/lksg/categories/general/subcategories/LksgProductionSpecific.kt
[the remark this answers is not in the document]
janherre: Done

dataland-frontend/src/components/forms/CreateLksgDataset.vue
```diff
@@ -153,15 +156,6 @@ export default defineComponent({
...
-        {
-          id: 0,
-          listOfGoodsOrServices: [] as string[],
-          allGoodsOrServicesAsString: "",
-        },
-      ],
-      idCounter: 0,
       waitingForData: false,
       dataDate: undefined as Date | undefined,
       companyAssociatedLksgData: {} as CompanyAssociatedDataLksgData,
```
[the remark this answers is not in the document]
janherre: These fields are removed now.

dataland-frontend/src/components/resources/frameworkDataSearch/lksg/LksgDataModel.ts
janherre: Implementing this without much boilerplate and not having unused arguments/variables would look like this:
`showIf: (dataModel: LksgData): boolean => !!dataModel`

This gets rid of the error, but is less understandable in my opinion.

dataland-frontend/tests/e2e/fixtures/lksg/LksgDataFixtures.ts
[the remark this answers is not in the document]
emanueldfine: Sorry I did not explain what we mean very well.
Siegfried and I think that the fake values for `capacity` should (also) contain actual words, because we expect the real-world-values to be smth like "2000 cars per hour" or "three tanks per month".

dataland-backend/src/main/kotlin/org/dataland/datalandbackend/model/lksg/categories/general/subcategories/LksgProductionSpecific.kt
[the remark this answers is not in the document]
emanueldfine: I just noticed that we could do the same kind of structuring also in the `sfdr` folder.
It also has the categories `environmental` and `social`, and also some `subcategories` to each one of them.
Would be nice in my opinion.

dataland-frontend/src/components/resources/frameworkDataSearch/lksg/LksgDataModel.ts
emanueldfine: I think there is a way with better readability:
We could just make the `showIf()` accept an optional param.  Then the IDE should not complain anymore, no matter if a param is passed or not:

image

Would this be ok for you?

dataland-frontend/src/components/forms/parts/fields/NumberFormField.vue
```diff
@@ -2,10 +2,10 @@
   <div class="form-field">
     <UploadFormHeader :name="displayName" :explanation="info" :is-required="required" />
     <FormKit
-      type="number"
+      type="text"
```
emanueldfine, line 5: Just out of curiosity:
Why have you done this?  Can't FormKit do validations on number-type form fields?

dataland-frontend/src/components/forms/parts/fields/PercentageFormField.vue
```diff
@@ -0,0 +1,27 @@
+<template>
+  <div class="form-field">
+    <UploadFormHeader :name="`${displayName} (%)`" :explanation="info" :is-required="required" />
+    <FormKit
+      type="text"
```
emanueldfine, line 5: Again just out of curiosity:  Can't FormKit accept an actual number for cases like this? Does that have downsides?

dataland-frontend/src/components/resources/frameworkDataSearch/lksg/LksgCompanyDataTable.vue
```diff
@@ -41,28 +41,28 @@
         :key="reportingPeriod.dataId"
       >
```
emanueldfine, line 42: Based on PrimeVue (https://primevue.org/datatable/) the `<DataTable> component does not have an attribute `responsiveLayout`.  Why is it set to "scroll" then?  Do I overlook smth. maybe?

dataland-frontend/src/components/resources/frameworkDataSearch/lksg/LksgCompanyDataTable.vue
```diff
@@ -41,28 +41,28 @@
         :key="reportingPeriod.dataId"
```
emanueldfine, line 41: Why have we set `reorderableColumns` to `true` for the `<DataTable>`? 
Has UX/UI requested us to enable this functionality?

dataland-frontend/src/components/resources/frameworkDataSearch/lksg/LksgCompanyDataTable.vue
```diff
@@ -41,28 +41,28 @@
...
+            <span v-else-if="data.kpiFormFieldComponent == 'PercentageFormField'">
+              {{ data[reportingPeriod.dataId] }} %</span
+            >
+            <span v-else-if="typeof data[reportingPeriod.dataId] === 'object' && data[reportingPeriod.dataId]?.value">
+              {{ data[reportingPeriod.dataId].value }}
+            </span>
+            <span v-else>{{ data[reportingPeriod.dataId] }} </span>
+          </template>
         </template>
       </Column>
```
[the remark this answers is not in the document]
emanueldfine: After fixing this, also the "v-for" in the second column should do smth like 
`v-for="columnIdentifier of columnIdentifierObjects"

dataland-frontend/src/components/resources/frameworkDataSearch/lksg/LksgCompanyDataTable.vue
```diff
@@ -41,28 +41,28 @@
...
+                v-if="
+                  data[reportingPeriod.dataId].length > 1 ||
+                  data[reportingPeriod.dataId].some((el) => typeof el === 'object')
+                "
+                @click="openModalAndDisplayValuesInSubTable(data[reportingPeriod.dataId], data.kpiLabel, data.kpiKey)"
+                class="link"
+                >Show "{{ data.kpiLabel }}"
+                <em class="material-icons" aria-hidden="true" title=""> dataset </em>
+              </a>
+              <span v-else> data[reportingPeriod.dataId][0] </span>
```
emanueldfine, line 56: It seems that in some cases arrays are not handled and displayed correctly.   Look at this case I had locally with the fake fixture company "Achkinadze-Kreb":

image

emanueldfine: Even though some locations are available, they are not displayed in the popup.

dataland-frontend/src/components/resources/frameworkDataSearch/lksg/LksgCompanyDataTable.vue
emanueldfine: I've found that attribute in two more components. Do you know why?

dataland-frontend/src/components/resources/frameworkDataSearch/lksg/LksgCompanyDataTable.vue
```diff
@@ -41,28 +41,28 @@
         :key="reportingPeriod.dataId"
       >
         <template #body="{ data }">
```
emanueldfine, line 43: Why does the #body of this `<Column>` have the value `{ data }`, and the one above (which renders the kpi field names and tooltips etc.) has the value `slotProps` ?
Can't they both be the same?

dataland-frontend/src/components/resources/frameworkDataSearch/KpiDataObject.ts
```diff
@@ -4,6 +4,7 @@ export interface KpiDataObject {
   kpiKey: string;
   kpiLabel: string;
   kpiDescription: string;
+  kpiFormFieldComponent: string;
 
   [index: string]: string;
```
emanueldfine, line 9: I know it is working fine, but is this really a good idea in terms of readability?

Why don't we just give `KpiDataObject` a field `dataId` and a field `content` or `value` ?

Wouldn't this be much more understandble for an external reader of the code?

dataland-frontend/src/components/resources/frameworkDataSearch/lksg/LksgCompanyDataTable.vue
```diff
@@ -41,28 +41,28 @@
...
+                >Show "{{ data.kpiLabel }}"
+                <em class="material-icons" aria-hidden="true" title=""> dataset </em>
+              </a>
+              <span v-else> data[reportingPeriod.dataId][0] </span>
+            </template>
+            <span v-else-if="data.kpiFormFieldComponent == 'PercentageFormField'">
+              {{ data[reportingPeriod.dataId] }} %</span
+            >
+            <span v-else-if="typeof data[reportingPeriod.dataId] === 'object' && data[reportingPeriod.dataId]?.value">
+              {{ data[reportingPeriod.dataId].value }}
```
emanueldfine, line 62: Why do you assume that an object could have a field `.value` ?   I mean where does this assumption come from and why is this explicitly handled here?

emanueldfine: Doesn't this mean, that all objects that we pass to an lksg dataset field, must have a field `value`, otherwise they will not be displayed at all?

dataland-frontend/src/components/resources/frameworkDataSearch/lksg/LksgCompanyDataTable.vue
emanueldfine, line 68: What is the third `<Column>` with the header "Impact Area" for?
It seems to me that it does not do anything, right?

dataland-frontend/src/components/resources/frameworkDataSearch/lksg/LksgPanel.vue
```diff
@@ -113,6 +113,7 @@ export default defineComponent({
         kpiKey: kpiKey,
         kpiLabel: kpiField?.label ? kpiField.label : kpiKey,
         kpiDescription: kpiField?.description ? kpiField.description : "",
+        kpiFormFieldComponent: kpiField?.component ?? "",
         [dataIdOfLksgDataset]: kpiValue,
```
emanueldfine, line 117: Until now, we have used the terminology/hierarchy:
`lksgDataSet => areas => subareas => kpis`

Now you have introduced the types `Category` and `Subcategory`.

Therefore I suggest we also use this terminology consistently, e.g. in the `convertLksgDataToFrontendFormat()`.
The loops should loop through categories and subcategories.
Maybe you can even type the respective objects in the for loops with the correct type, e.g. the current `subAreaObject` could be of type `Subcategory` ?

dataland-frontend/src/components/resources/frameworkDataSearch/lksg/LksgPanel.vue
```diff
@@ -113,6 +113,7 @@ export default defineComponent({
...
       } as KpiDataObject;
```
emanueldfine, line 118: I think in `createKpiDataObjects()` the writing to the map is a little bit misleading.

I suggest just doing this instead of the current last 4 lines of code:

`const potentiallyExistingKpiMapEntry = this.kpiDataObjects.get(kpiKey);
      if (potentiallyExistingKpiMapEntry ) Object.assign(potentiallyExistingKpiMapEntry , kpiData);
      else this.kpiDataObjects.set(kpiKey, kpiData);`

emanueldfine: Furthermore I find the name of `kpiDataObjects` a little bit misleading.
I'd prefer if I we name it in a way, that you see it is map right away.
Maybe smth. like `mapOfKpiKeysToKpiDataObjects`

dataland-frontend/src/components/resources/frameworkDataSearch/lksg/LksgPanel.vue
```diff
@@ -113,6 +113,7 @@ export default defineComponent({
...
       let existingKpi = this.kpiDataObjects.get(kpiKey);
```
emanueldfine, line 119: The method `sortReportingPeriodsToDisplayAsColumns()` has a strange name for its input param.  It says `listOfDataDateToDisplayAsColumns`, even though the list contains column identifier objects.

dataland-frontend/src/components/resources/frameworkDataSearch/KpiDataObject.ts
emanueldfine: Ok now I see.   Due to the way we build the map in the `LksgPanel.vue` it is not possible at the moment.
But it would be possible if a `KpiDataObject` would have a Map with dataId to value!
We could discuss this if you like.

dataland-backend/src/main/kotlin/org/dataland/datalandbackend/model/lksg/categories/general/subcategories/LksgProductionSpecific.kt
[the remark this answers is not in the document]
SiegfriedSobkowiak: But it is kind of out of scope for this story. Maybe make a note to the next SFDR story in the backlog?

dataland-frontend/src/components/resources/frameworkDataSearch/lksg/LksgDataModel.ts
SiegfriedSobkowiak: I like the idea

dataland-frontend/src/components/forms/parts/fields/ProductionSitesFormField.vue
```diff
@@ -34,19 +41,12 @@ import UploadFormHeader from "@/components/forms/parts/elements/basic/UploadForm
...
-      listOfGoodsOrServicesString: "",
-      listOfProductionSites: [
-        {
-          id: 0,
-          listOfGoodsOrServices: [] as string[],
-          listOfGoodsOrServicesString: "",
-        },
-      ],
+      existingProductionSites: [] as LksgProductionSite[],
+      listOfProductionSiteIds: [0] as number[],
```
emanueldfine, line 49: Why do we need to initialize this list with "0" as already existing production site id?

dataland-frontend/tests/e2e/fixtures/lksg/LksgDataFixtures.ts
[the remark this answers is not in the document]
SiegfriedSobkowiak: changed

dataland-backend/src/main/kotlin/org/dataland/datalandbackend/model/lksg/categories/general/subcategories/LksgProductionSpecific.kt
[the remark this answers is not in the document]
emanueldfine: Agree in general, but in this case with IntelliJ it should take only a minute to refactor that. The IDE takes care of the rest.

Of course, if there is more trouble to it, let's postpone it.

dataland-frontend/src/components/resources/frameworkDataSearch/lksg/LksgDataModel.ts
SiegfriedSobkowiak: I do not share that concern. Maybe we can discuss this briefly.

dataland-backend/src/main/kotlin/org/dataland/datalandbackend/model/lksg/categories/general/subcategories/LksgProductionSpecific.kt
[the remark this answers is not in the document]
SiegfriedSobkowiak: done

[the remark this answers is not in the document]
emanueldfine: Nice, thanks.
Could you create a note (in the ticket for the sfdr form creation maybe?) that we also put the `SfdrGeneral` sub-model directly into the `SfdrData` main model, instead of nesting it under `SfdrSocial` ?
This would then be consistent with Lksg, and I also think it makes sense.
What do you think?

dataland-frontend/src/components/resources/frameworkDataSearch/lksg/LksgPanel.vue
SiegfriedSobkowiak: changed

dataland-frontend/src/components/forms/parts/fields/NonNegativeNumberFormField.vue
```diff
@@ -0,0 +1,26 @@
...
+import { defineComponent } from "vue";
+import { FormKit } from "@formkit/vue";
+import { FormFieldPropsWithPlaceholder } from "@/components/forms/parts/fields/FormFieldProps";
+
+export default defineComponent({
+  name: "NonNegativeNumberFormField",
+  components: { FormKit, UploadFormHeader },
+  props: FormFieldPropsWithPlaceholder,
+});
+</script>
```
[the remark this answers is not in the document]
MarcTM01: Should be possible. Will take care of it

dataland-frontend/src/components/forms/parts/fields/NumberFormField.vue
MarcTM01: Not always. Let's talk about this.

dataland-frontend/src/components/resources/frameworkDataSearch/lksg/LksgDataModel.ts
emanueldfine: Ok we can talk about it.
Another point I don't like about it, is that it makes you think that it is just hidden if `showIf()` returns `false`, while in fact it is not rendered at all in this case.

dataland-frontend/src/components/forms/parts/fields/PercentageFormField.vue
```diff
@@ -0,0 +1,27 @@
...
+      :name="name"
+      :validation-label="validationLabel ?? displayName"
+      :validation="`number|between:0,100|${validation}`"
+      :placeholder="placeholder ?? 'Value in %'"
+      inner-class="short"
+      v-model="percentageString"
```
emanueldfine, line 11: I could not find out what variable this is referring to.
I mean if I search for "percentageString" in our frontend source code, I cannot find any references.
Where does this come from?

dataland-frontend/src/components/forms/parts/fields/NonNegativeNumberFormField.vue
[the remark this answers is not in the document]
emanueldfine: There are more generic field components that are almost copies of each other.
Maybe we should go through them together and think about the same approach there:
Instead of two seperate components, just use one, and adjust the "validation" or "name" based on the use case.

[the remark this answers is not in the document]
emanueldfine: E.g. `FreeTextFormField` and `InputTextFormField` have only one diff:   One has `type` set to "text", and the other one to "textarea".
If we create a generic `TextFormField` and pass the type as prop, we could avoid this.

dataland-frontend/src/components/resources/frameworkDataSearch/lksg/LksgCompanyDataTable.vue
fsoffel: I think this is for filtering data points. They are of a type which has it's actual value stored in a field called value.

dataland-frontend/src/components/forms/parts/fields/ProductionSitesFormField.vue
emanueldfine: Got it, thx.

dataland-frontend/src/components/forms/parts/fields/NonNegativeNumberFormField.vue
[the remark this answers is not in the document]
MarcTM01: Fixed

dataland-frontend/src/components/forms/parts/fields/PercentageFormField.vue
MarcTM01: Removed.

dataland-frontend/src/components/forms/CreateLksgDataset.vue
```diff
@@ -123,8 +125,6 @@ export default defineComponent({
   },
```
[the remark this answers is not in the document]
MarcTM01: Will look into this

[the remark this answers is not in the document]
MarcTM01: Fields are now consistently called `label` or `description`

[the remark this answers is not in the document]
emanueldfine: Perfect. Thx

dataland-frontend/src/components/forms/parts/fields/PercentageFormField.vue
emanueldfine: Ok. Thanks

dataland-frontend/src/components/resources/frameworkDataSearch/lksg/LksgCompanyDataTable.vue
emanueldfine: Ok if that is the case then...

1.) The type check is not sufficient. It checks if the kpiValue is of type "object". But it is not guaranteed that each "object" type kpiValue is of type DataPoint.  

2.) The Lksg model does not even have a DataPoint-type field.   So it does not make sense in the first place.

My point is that I think this code is dead and instable, because it does an assumption (it assumes that all "object" type kpiValues have a field ".value")

fsoffel: It is not assumed that all objects have type DataPoint. That is why the existence of the `value` field is checked.
Ad 2: Hmmm. Yes. Maybe my guess was wrong. However, this is how we used it in the document download branch.

dataland-frontend/src/components/resources/frameworkDataSearch/KpiDataObject.ts
janherre: The main reason we construct this kind of objects, is because we need to provide an array of these objects to the `DataTable` component. I do not know if `DataTable` can handle a map. We would need to investigate.

dataland-frontend/src/components/resources/frameworkDataSearch/lksg/LksgCompanyDataTable.vue
janherre: The second check is to ensure the object has a field `value`

janherre: `data[reportingPeriod.dataId]?.value` is falsy if .value returns undefined

dataland-frontend/src/components/forms/parts/fields/NonNegativeNumberFormField.vue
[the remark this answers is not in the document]
emanueldfine: The text form fields cannot be easily united, since the `type` prop cannot be passed in the `v-for <component/>`.
Therefore we leave it as it is.

We could not find another pair of components, that could be united to a more generic component.

Therefore I close this.

dataland-frontend/src/components/resources/frameworkDataSearch/KpiDataObject.ts
emanueldfine: No you would still provide an array to `<DataTable>`.   But the objects in that array would have a field which contains a map of dataId to the value.

We can discuss this if you like.

dataland-frontend/src/components/forms/parts/fields/NonNegativeNumberFormField.vue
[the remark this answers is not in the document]
fsoffel: What do you mean by the `type` prop cannot be passed in the `<component/>`? It sure would need to be added in the lksg converter, but why would that not be possible. I prefer the current approach though.

dataland-frontend/src/components/resources/frameworkDataSearch/lksg/LksgCompanyDataTable.vue
emanueldfine: Ah ok.  Then 1) is answered. Thanks. I overlook that.

But still this is dead code currently.   Should we therefore remove it?
What do you think?

dataland-frontend/src/components/forms/parts/fields/NonNegativeNumberFormField.vue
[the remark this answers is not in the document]
emanueldfine: Sorry I did not describe it precisely.
Yes, `type` could be added, and then we could introduce a generic Text-component, but Marc had some good reasons why it might be better to not do it at the moment.

dataland-frontend/src/components/resources/frameworkDataSearch/lksg/LksgCompanyDataTable.vue
emanueldfine: Well in sfdr `DataPointBigDecimal` and `DataPointYesNo` are actually being used.  Both are objects with a `value` field.
There the code would not be dead.

Maybe it was copy pasted?

dataland-frontend/src/components/resources/frameworkDataSearch/lksg/LksgCompanyDataTable.vue
janherre: This is a bug with how the headers are generated. I fixed it for this specific case by removing the header completely since it is kind of redundant for the single column "Countries" table. For all other cases it needs to be made sure that there is a correct specification for the headers in `DataModelsTranslations.ts`, which is rather inconvenient, I admit, so if you have a better idea we could discuss.

dataland-frontend/src/components/resources/frameworkDataSearch/lksg/LksgCompanyDataTable.vue
janherre: It is part of the rowGroupHeaders. Otherwise the grey header bar does not span over the whole width of the table. The header is unnecessary and is removed.

dataland-frontend/src/components/resources/frameworkDataSearch/lksg/LksgCompanyDataTable.vue
janherre: Changed to consistent usage of `slotProps` as that enables the typescript server to recognize the allowed fields correctly (we could also use `data` directly to make it shorter but then everything is marked by the linter)

dataland-frontend/src/components/resources/frameworkDataSearch/lksg/LksgCompanyDataTable.vue
janherre: It seems to have been removed with a primevue update, the IDE thinks this exists, but there is no such prop in the code or in the docs anymore. Removing it hasn't changed anything. Therefore I suggest to remove it.

dataland-frontend/src/components/resources/frameworkDataSearch/lksg/LksgCompanyDataTable.vue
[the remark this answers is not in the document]
janherre: I have refactored the very generic `columnIdentifierObjects` it with the introduction of a type `DataSetReportingPeriod` that is more specific and makes more sense here in my opinion. The same goes for the following reviews that deal with this variable.

dataland-frontend/src/components/resources/frameworkDataSearch/lksg/LksgPanel.vue
janherre: refactored to the type `DataSetReportingPeriod` , see above.

dataland-frontend/src/components/forms/parts/fields/PercentageFormField.vue
emanueldfine: Broken: In Firefox I can type "L" into the "representedEmployees" field and submit that.
It is NOT actually submitted, but looks like it to the user (Successful Upload badge!)

dataland-frontend/src/components/resources/frameworkDataSearch/lksg/LksgCompanyDataTable.vue
emanueldfine: It is indeed possible, I've tried it locally. But I don't think that we ever explicitily decided that we want this.
Therefore it is ok that you took it away and I close this conversation.

dataland-frontend/src/components/resources/frameworkDataSearch/lksg/LksgCompanyDataTable.vue
emanueldfine: It is still there for `FrameworkDataSearchResults.vue`

dataland-frontend/src/components/resources/frameworkDataSearch/lksg/LksgCompanyDataTable.vue
emanueldfine: Nice, thx.

dataland-backend/src/main/kotlin/org/dataland/datalandbackend/model/lksg/categories/general/subcategories/LksgProductionSpecific.kt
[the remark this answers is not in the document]
SiegfriedSobkowiak: Agreed, I put a comment into the SFDR refactoring story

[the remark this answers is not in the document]
emanueldfine: Perfect. Thanks. Closing this.

dataland-frontend/src/components/resources/frameworkDataSearch/lksg/LksgPanel.vue
emanueldfine: Let's meet halfway and call it `mapOfKpiKeysToDataObjects`, it's only slightly longer than your suggestion.

I agree that a shorter name would be better to read, but in this case I had really no clue at all what this map is about (until I consoled.logged it), and I think that name would help other developers to understand it better.

dataland-frontend/src/components/resources/frameworkDataSearch/lksg/LksgCompanyDataTable.vue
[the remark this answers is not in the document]
emanueldfine: Having a type for the column identifier objects is nice, I like that.

But the naming of that type with `DataSetReportingPeriod` is the opposite of what I meant:

I actually liked the idea of calling it `ColumnIdentifierObject` or something similar.
`DataSetReportinPeriod` is misleading because it implies that it only bears a reporting period, while it also bears the data Id and serves the sole purpose of identifying the different columns.

The current naming leads to code which is strange to read, like here:

image

The first time I read this I was really confused.

dataland-frontend/src/components/resources/frameworkDataSearch/lksg/LksgCompanyDataTable.vue
emanueldfine: Thanks, I will look into this and reply as soon as I have made up my mind about it.

dataland-frontend/src/components/resources/frameworkDataSearch/lksg/LksgPanel.vue
emanueldfine: I replied to the refactoring with a small change request in the conversation above.

Will keep this conversation here also open until the one above is closed.

dataland-frontend/src/components/resources/frameworkDataSearch/lksg/LksgCompanyDataTable.vue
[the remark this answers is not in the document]
janherre: The reason I did not want to name it `columnIdentifierObject` is because I find it very confusing when such an abstract object has a very concrete field named `reportingPeriod`, there is no logical connection between the object name and the field name. So I would go with your second suggestion `DataSetReportingPeriodAndDataId` . Additionally I would refactor this loop to work like `v-for="(reportingPeriod, dataId) of dataSetReportingPeriodAndDataId"`. This is a much clearer way to extract this tuple, is that alright for you?

dataland-frontend/src/components/resources/frameworkDataSearch/lksg/LksgCompanyDataTable.vue
janherre: Removed it

dataland-frontend/src/components/resources/frameworkDataSearch/lksg/LksgCompanyDataTable.vue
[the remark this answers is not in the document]
emanueldfine: Sounds great, thx !

dataland-frontend/src/components/resources/frameworkDataSearch/lksg/LksgPanel.vue
janherre: Renamed it to `mapOfKpiKeysToDataObjects`

dataland-frontend/src/components/resources/frameworkDataSearch/lksg/LksgCompanyDataTable.vue
emanueldfine: 👍

dataland-frontend/src/components/resources/frameworkDataSearch/lksg/LksgPanel.vue
emanueldfine: I think this broke the CI because in the `LksgPanel.vue` the prop of `<LksgCompanyDataTable>` still says `:kpiDataObjects`, even though the actual prop of that component was renamed to `mapOfKpiKeysToDataObjects`.

However, I just noticed that `LksgCompanyDataTabe` does not even need the complete `mapOfKpiKeysToDataObjects`, because it only extracts the values from that map and does not even use the keys!
Therefore we can also change that prop to an array of `KpiDataObject`, and rename it to `kpiDataObjects`, and then just pass the values of `mapOfKpiKeysToDataObjects` in the `LskgPanel.vue` down to `<LksgCompanyDataTable/>`

dataland-frontend/src/components/resources/frameworkDataSearch/lksg/LksgCompanyDataTable.vue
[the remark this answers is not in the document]
emanueldfine: Yes it has, thank you.  There's only one thing that still needs to be adjusted:

It is the prop name of the array, which holds the `DataSetReportingPeriodAndDataId`-objects.
It is still named in a way, that one might think it only holds reporting periods:

image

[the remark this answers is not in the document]
emanueldfine: And why have you added an index signature to `ReportingPeriodOfDataSetWithId`?

image

dataland-frontend/src/components/resources/frameworkDataSearch/lksg/LksgPanel.vue
janherre: You are right, this is fixed and changed to an array.

dataland-frontend/src/components/resources/frameworkDataSearch/KpiDataObject.ts
janherre: I have tried to change the field `[dataId]: kpiValue` to a map like `content: {[dataId]: kpiValue}` . This unfortunately breaks the table for more than one reporting period, the adding of data for more than one reportingPeriod to the table is handled with the `Object.assign()` method which is a bit tricky to handle. I will give this one to Benedikt to investigate.

janherre: I managed to push a working version. Still Benedikt is the one who knows about this and can continue if something is broken now.

dataland-frontend/src/components/resources/frameworkDataSearch/lksg/LksgDataModel.ts
emanueldfine: Agreed on not renaming

dataland-frontend/src/components/resources/frameworkDataSearch/lksg/LksgCompanyDataTable.vue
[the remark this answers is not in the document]
emanueldfine: Perfect. Closing this.

dataland-frontend/src/components/resources/frameworkDataSearch/lksg/LksgPanel.vue
emanueldfine: Problem was fixed in the conversation above, therefore closing this one here too

dataland-frontend/src/components/resources/frameworkDataSearch/lksg/LksgPanel.vue
emanueldfine: Works. Thanks.

dataland-frontend/src/components/resources/frameworkDataSearch/lksg/LksgPanel.vue
emanueldfine: Discussed with Siegfried and adjusted.

dataland-frontend/src/components/resources/frameworkDataSearch/lksg/LksgCompanyDataTable.vue
emanueldfine: Florian and I decided to not remove this code. It is true that it is currently not executed, but it assures that actual DataPoint objects can be displayed, so it kind of makes this template more "stable" and "safe" for future changes of the data model.

dataland-frontend/src/components/resources/frameworkDataSearch/lksg/LksgCompanyDataTable.vue
```diff
@@ -111,7 +127,7 @@ export default defineComponent({
     },
   },
   mounted() {
-    this.kpiDataObjectsToDisplay = Array.from(this.kpiDataObjects.values());
+    this.kpiDataObjectsToDisplay = this.arrayOfKpiDataObjects;
```
emanueldfine, line 130: `kpiDataObjectsToDisplay` is pointless.  It just bears the value of the prop `arrayOfKpiDataObjects`, and also causes a type violation here, because 
`kpiDataObjectsToDisplay` is of type `Array<KpiDataObject>`
and
`arrayOfKpiDataObjects` is of type `Array<KpiDataObject> | undefined`

So why not delete `kpiDataObjectsToDisplay`?

dataland-frontend/src/components/resources/frameworkDataSearch/lksg/LksgCompanyDataTable.vue
emanueldfine: 1.) Why don't we rename `tableType` to `kpiKeyOfTable` ?   At the end this just contains the kpi field name...

emanueldfine: 2.) Why the watcher for `listOfRowContents` ?  This should not change anymore..

emanueldfine: 3.) In `generateColsNames()`, why do you check if `listOfRowContents` is an array? It already is defined as array, so why checking that?

emanueldfine: 4.) In `generateColumns()` we set the same value `key` in both fields of the object that we `push` into `columns`.
  So why does it even have two fields in the first place?

emanueldfine: 7.) Typo in `humanizeStringIfNeccessary()`

emanueldfine: 8.) This whole component only works for objects which have only one sub-level.  Shouldn't we therefore use the power of typescript and give the `listOfRowContents` a type which reflects that?

emanueldfine: 8.) I think the component can only handle two-level objects.  Shouldn't we use the power of typescript and reflect that in the type of the prop?

e.g. `type TwoLevelObject = {
  [key: string]: {
    [subKey: string]: any;
  };
};`

dataland-frontend/src/components/resources/frameworkDataSearch/lksg/LksgCompanyDataTable.vue
BenediktLaubmann: done

dataland-frontend/src/components/resources/frameworkDataSearch/lksg/LksgCompanyDataTable.vue
BenediktLaubmann: If I remember correctly, the original purpose of the DetailsCompanyDataTable component has been to display the production sites. We chose to generalize this component in order to display several subtables including list of nace codes or list of country codes or any other lists or list of objects like production sites. This may be the case why it appears complicated.

BenediktLaubmann: 1) sounds fine to me. I rename it

dataland-frontend/src/components/resources/frameworkDataSearch/lksg/LksgCompanyDataTable.vue
emanueldfine: Thanks. I just noticed that the prop `tableDataTitle` is never used. Could you delete it?

Same goes for that prop with the same name in the file `CompanyDataTable.vue`

dataland-frontend/src/components/resources/frameworkDataSearch/lksg/LksgCompanyDataTable.vue
BenediktLaubmann: 2. I don't know - I would put it into mounted, then it will run only once in the beginning as expected

BenediktLaubmann: 3) & 5) : these queries are indeed deprecated. I remove them

emanueldfine: 10.) `columnHeaders: detailsCompanyDataTableColumnHeaders`    in the `data{}` object is unnecessary. You could directly write `detailsCompanyDataTableColumnHeaders` there, and also use this in the HTML template.
This also helps developers, because if the STRG-click on it to check what this is, they get redirected to that `const` right away.

emanueldfine: Note: I deleted 8) because you kind of handle multi-level objects.   You display their `.value` if they have that field on the first level, otherwise you directly display the object (which will look weird if it actually happens, but since this component is generic, I think it is ok for now)

BenediktLaubmann: > 4.) In `generateColumns()` we set the same value `key` in both fields of the object that we `push` into `columns`. So why does it even have two fields in the first place?

simplified it by reducing the columns from list of object with two identical strings to a list of just this string (the columns name/label/header)

BenediktLaubmann: > 2.) Why the watcher for `listOfRowContents` ? This should not change anymore..

I don't know, I will put its actions (the function generateColsNames()) into mounted, then it will just execute once in the beginning as desired

BenediktLaubmann: > 11.) Small renaming request. `columns` is a little misleading, since it only contains the keys of fields, whose values shall be displayed in a column. So I would suggest we rename it to smth. like `keysOfValuesForColumnDisplay` and then `v-for="keyOfValue of keysOfValuesForColumnDisplay"` in the template.

done, but with `"keyOfColumn"` instead of `"keyOfValue"` which sounds a little bit too generic to me

BenediktLaubmann: > 6.) In the `mounted` block, there is already an `if`-statement which checks if the received array is an array of strings. You could directly set the `columns` array to `[{ field: this.tableType, header: this.tableType }]` there, and not execute the `generateColumns()` anymore, since it is pointless at this point. Then you wouldn't need the mapping `this.listOfRowContents = dialogRefData.listOfRowContents.map((o) => ({ [this.tableType]: o }));` anymore.

I am not sure if this works as we expect. Lets have a quick talk about it.

BenediktLaubmann: > 6.) In the `mounted` block, there is already an `if`-statement which checks if the received array is an array of strings. You could directly set the `columns` array to `[{ field: this.tableType, header: this.tableType }]` there, and not execute the `generateColumns()` anymore, since it is pointless at this point. Then you wouldn't need the mapping `this.listOfRowContents = dialogRefData.listOfRowContents.map((o) => ({ [this.tableType]: o }));` anymore.

Turns out we need the mapping here, because otherwise, the contents for the case of only one column won't get displayed.

dataland-frontend/src/components/resources/frameworkDataSearch/lksg/LksgCompanyDataTable.vue
BenediktLaubmann: done, waiting for green CI

emanueldfine: It is still there as set prop in some other components.  Search fo `tableDataTitle` in the `src` folder of `dataland-frontend` to see it.

emanueldfine: The CI is failing because of this statement in the component-test of a component that used to have `tableDataTitle`

`expect(wrapper.vm.tableDataTitle).to.be.a("string").that.is.empty`

dataland-frontend/src/components/resources/frameworkDataSearch/lksg/LksgCompanyDataTable.vue
emanueldfine: > > 6.) In the `mounted` block, there is already an `if`-statement which checks if the received array is an array of strings. You could directly set the `columns` array to `[{ field: this.tableType, header: this.tableType }]` there, and not execute the `generateColumns()` anymore, since it is pointless at this point. Then you wouldn't need the mapping `this.listOfRowContents = dialogRefData.listOfRowContents.map((o) => ({ [this.tableType]: o }));` anymore.
> 
> Turns out we need the mapping here, because otherwise, the contents for the case of only one column won't get displayed.

Yes you are right. Sorry about that comment.   But you still can set `keysOfValuesForColumnDisplay` directly to `[ this.tableType ]` in the `if`-statement, and only execute `generateColumns()` in the `else` block.

dataland-frontend/src/components/resources/frameworkDataSearch/lksg/LksgCompanyDataTable.vue
emanueldfine: Fixed. Closing it.

dataland-frontend/src/components/resources/frameworkDataSearch/lksg/LksgCompanyDataTable.vue
emanueldfine: Perfect! Thanks for all the fixes. I am closing this now since everything has been addressed!

PR #232, opened 2023-05-22
Dala 1512 removing kpi section after adding leads to error on request 2

fsoffel:
# Pull Request \<Title>
`<Description here>`
- [ ] A peer-review has been executed
- [ ] Documentation is updated as required
- [ ] The automated deployment is updated if required
- [ ] The new version is deployed with fake data to the dev server "dev1" using this branch
  - [ ] Run with setting `Use real data` turned off and `Reset non-user related Docker Volumes & Re-populate` turned on 
  - [ ] The new feature is manually used/tested/observed on dev server
  - [ ] All implemented Social Logins have been tested manually in the UI
- [ ] The new version is deployed with real data to a dev server using this branch
  - [ ] Run with both settings `Reset non-user related Docker Volumes & Re-populate` and `Use real data` turned on
- [ ] If any work on the UI is to be merged, those changes were also documented in the Figma


dataland-frontend/src/components/forms/CreateEUTaxonomyForFinancials.vue
fsoffel, line 1: I expect the edit functionality not to work because there was no change in `fetchTemplateData` to convert the API response to the new model.

dataland-frontend/src/components/forms/CreateEUTaxonomyForFinancials.vue
```diff
@@ -273,8 +248,36 @@
                         </div>
                       </div>
                     </FormKit>
-                  </FormKit>
-                </div>
+
+                    <FormKit :name="financialServiceOption.value" type="group">
```
fsoffel, line 252: Why is this not the same formkit group as its upper sibling. They have the same name.

dataland-frontend/src/components/forms/CreateEUTaxonomyForFinancials.vue
```diff
@@ -552,13 +557,39 @@ export default defineComponent({
         this.postEuTaxonomyDataForFinancialsProcessed = false;
         this.messageCount++;
 
+        // JSON.parse/stringify used to clone the formInputsModel in order to stop Proxy refreneces
```
fsoffel, line 560: I find you should put this added model conversion code into a function for the sake of readability

dataland-frontend/src/components/forms/CreateEUTaxonomyForFinancials.vue
```diff
@@ -552,13 +557,39 @@ export default defineComponent({
...
+            const fieldName = euTaxonomyKPIsModel.kpisFieldNameToFinancialServiceType[kpiKey];
+            if(kpiKey){ delete kpi[kpiKey][fieldName] };
+            return kpi;
+          })
+          .reduce( (all, one) => ({ ...all, ...one }));
+
+        delete (clonedFormInputsModel.data as ObjectType).kpiSectionsModel;
+        clonedFormInputsModel.data = {...clonedFormInputsModel.data as ObjectType, eligibilityKpis, ...kpis};
+
+        console.log(clonedFormInputsModel);
```
fsoffel, line 583: TODO: these two lines must be removed

dataland-frontend/src/components/forms/CreateEuTaxonomyForFinancials.vue
```diff
@@ -437,6 +440,7 @@ export default defineComponent({
       message: "",
       namesOfAllCompanyReportsForTheDataset: [] as string[],
       templateDataset: undefined as undefined | EuTaxonomyDataForNonFinancials,
+      kpiSections: []
```
fsoffel, line 443: this is not used

dataland-frontend/src/components/forms/CreateEuTaxonomyForFinancials.vue
```diff
@@ -554,14 +580,37 @@ export default defineComponent({
...
         this.postEuTaxonomyDataForFinancialsProcessed = false;
         this.messageCount++;
+        
+        // JSON.parse/stringify used to clone the formInputsModel in order to stop Proxy refreneces
+        const clonedFormInputsModel = JSON.parse(JSON.stringify(this.formInputsModel)) as ObjectType;
+        const kpiSectionsModel = (clonedFormInputsModel.data as ObjectType).kpiSectionsModel;
+
+        const eligibilityKpis = Object.keys(kpiSectionsModel)
+          .map( key => ({ key, fieldName: euTaxonomyKPIsModel.kpisFieldNameToFinancialServiceType[key] as string }) )
+          .map( kpi => ({ [kpi.fieldName]: kpiSectionsModel[kpi.key][kpi.fieldName] }) )
```
fsoffel, line 590: if `kpi.fieldName` is `undefined` this property just does not get set?

dataland-frontend/src/components/forms/CreateEuTaxonomyForFinancials.vue
```diff
@@ -554,14 +580,37 @@ export default defineComponent({
...
+          .reduce( (all, one) => ({ ...all, ...one }));
+
+        const kpis = Object.keys(kpiSectionsModel)
+          .filter(key => key !== 'assetManagementKpis')
+          .map( key => ({ [key]: kpiSectionsModel[key] }))
+          .map(kpi => {
+            const kpiKey = Object.keys(kpi)[0];
+            const fieldName = euTaxonomyKPIsModel.kpisFieldNameToFinancialServiceType[kpiKey];
```
fsoffel, line 598: it should not be named field name but rather financialServiceType. Same for line 589

dataland-frontend/src/components/forms/CreateEuTaxonomyForFinancials.vue
```diff
@@ -554,14 +580,37 @@ export default defineComponent({
       try {
         this.postEuTaxonomyDataForFinancialsProcessed = false;
         this.messageCount++;
+        
+        // JSON.parse/stringify used to clone the formInputsModel in order to stop Proxy refreneces
+        const clonedFormInputsModel = JSON.parse(JSON.stringify(this.formInputsModel)) as ObjectType;
+        const kpiSectionsModel = (clonedFormInputsModel.data as ObjectType).kpiSectionsModel;
+
+        const eligibilityKpis = Object.keys(kpiSectionsModel)
```
fsoffel, line 588: I find these both filterings happening here too complicated. Why not use one call to filter in which we check that the key is contained in a predefined array of expected values.

dataland-frontend/src/components/forms/CreateEuTaxonomyForFinancials.vue
```diff
@@ -506,7 +510,29 @@ export default defineComponent({
               const receivedFormInputsModel = convertValuesFromDecimalsToPercentages(
                 companyAssociatedEuTaxonomyData as ObjectType
               );
+
+              const kpiSectionsModel = Object.keys(euTaxonomyKPIsModel.kpisFieldNameToFinancialServiceType)
+                .map( key => {
+                  const fieldName = euTaxonomyKPIsModel.kpisFieldNameToFinancialServiceType[key];
+                  const kpi = (receivedFormInputsModel.data as ObjectType).eligibilityKpis[fieldName];
```
fsoffel, line 517: `kpi` is wrong naming. Basically a kpi corresponds to a single data point. `kpis` would fit. Better `eligibilityKpis`.

dataland-frontend/src/components/forms/CreateEuTaxonomyForFinancials.vue
```diff
@@ -506,7 +510,29 @@ export default defineComponent({
               const receivedFormInputsModel = convertValuesFromDecimalsToPercentages(
                 companyAssociatedEuTaxonomyData as ObjectType
               );
+
+              const kpiSectionsModel = Object.keys(euTaxonomyKPIsModel.kpisFieldNameToFinancialServiceType)
+                .map( key => {
+                  const fieldName = euTaxonomyKPIsModel.kpisFieldNameToFinancialServiceType[key];
```
fsoffel, line 516: As below: don't call it field name

dataland-frontend/src/components/forms/CreateEuTaxonomyForFinancials.vue
```diff
@@ -506,7 +510,29 @@ export default defineComponent({
...
+                  const kpi = (receivedFormInputsModel.data as ObjectType).eligibilityKpis[fieldName];
+                  const eligibilityKpi = kpi ? {[fieldName]: kpi } : null;
```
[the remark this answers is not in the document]
fsoffel: Also undefined > null. But that might be personal preference. If this works, I'm fine.

dataland-frontend/src/components/forms/CreateEuTaxonomyForFinancials.vue
```diff
@@ -506,7 +510,29 @@ export default defineComponent({
...
+                  return { [key]: eligibilityKpi };
+                })
+                .map( item => {
+                  const key = Object.keys(item)[0];
```
fsoffel, line 522: financialServiceType > key

dataland-frontend/src/components/forms/CreateEuTaxonomyForFinancials.vue
```diff
@@ -506,7 +510,29 @@ export default defineComponent({
...
               );
+
+              const kpiSectionsModel = Object.keys(euTaxonomyKPIsModel.kpisFieldNameToFinancialServiceType)
+                .map( key => {
+                  const fieldName = euTaxonomyKPIsModel.kpisFieldNameToFinancialServiceType[key];
+                  const kpi = (receivedFormInputsModel.data as ObjectType).eligibilityKpis[fieldName];
+                  const eligibilityKpi = kpi ? {[fieldName]: kpi } : null;
+                  return { [key]: eligibilityKpi };
+                })
+                .map( item => {
```
fsoffel, line 521: Why two different map calls?

dataland-frontend/src/components/forms/CreateEuTaxonomyForFinancials.vue
```diff
@@ -506,7 +510,29 @@ export default defineComponent({
...
+                  const key = Object.keys(item)[0];
+                  const kpi = receivedFormInputsModel.data[key];
```
fsoffel, line 523: kpis > kpi

dataland-frontend/src/components/forms/CreateEuTaxonomyForFinancials.vue
```diff
@@ -506,7 +510,29 @@ export default defineComponent({
...
+                  if(kpi){
+                    item[key] = {...item[key], ...receivedFormInputsModel.data[key]};
```
fsoffel, line 525: You could use kpi here instead of receivedFormInputsModel.data[key]

dataland-frontend/tests/e2e/utils/EuTaxonomyFinancialsUpload.ts
```diff
@@ -273,3 +273,40 @@ export function uploadCompanyViaApiAndEuTaxonomyDataForFinancialsViaForm(
...
+ * via the form in the frontend, and then visits the view page where that dataset is displayed
+ * @param companyInformation Company information to be used for the company upload
+ * @param testData EU Taxonomy dataset for financial companies to be uploaded
+ * @param beforeFormFill is performed before filling the fields of the upload form
+ * @param formFillSteps Steps involved to fill data of the upload form
+ * @param afterFormFill is performed after filling the fields of the upload form
+ * @param submissionDataIntercept performs checks on the request itself
+ * @param afterDatasetSubmission is performed after the data has been submitted
+ */
+export function companyViaApiAndEuTaxonomyDataForFinancialsViaForm(
```
fsoffel, line 288: Function name should include what it does, i.e. upload.

fsoffel, line 288: If you want to keep it here, it's origin must be deleted

dataland-frontend/src/components/forms/CreateEuTaxonomyForFinancials.vue
```diff
@@ -201,70 +201,45 @@
...
-                      class="p-button-text ml-auto"
-                      icon="pi pi-trash"
-                    ></PrimeButton>
-                  </div>
-
-                  <FormKit
-                    v-if="financialServiceOption.value !== 'assetManagementKpis'"
-                    :name="financialServiceOption.value"
-                    type="group"
+                <FormKit v-model="kpiSections" name="kpiSectionsModel" type="group">
```
fsoffel, line 204: The name kpiSectionsModel does not seem fit to me. I don't get the Model suffix.

dataland-frontend/tests/e2e/specs/eu-taxonomy-financials/AddRemoveKpis.ts
```diff
@@ -0,0 +1,113 @@
...
+
+    before(function () {
+      cy.fixture("CompanyInformationWithEuTaxonomyDataForFinancialsPreparedFixtures").then(function (jsonContent) {
+        const preparedFixtures = jsonContent as Array<FixtureData<EuTaxonomyDataForFinancials>>;
+        testData = getPreparedFixture("company-for-all-types", preparedFixtures);
+      });
+    });
+
+    it("Check", () => {
+      testData.companyInformation.companyName = "company-for-all-types";
```
fsoffel, line 33: You could set a different name which makes the company's origin clearer

dataland-frontend/tests/e2e/specs/eu-taxonomy-financials/AddRemoveKpis.ts
```diff
@@ -0,0 +1,113 @@
...
+    let testData: FixtureData<EuTaxonomyDataForFinancials>;
+
+    before(function () {
+      cy.fixture("CompanyInformationWithEuTaxonomyDataForFinancialsPreparedFixtures").then(function (jsonContent) {
+        const preparedFixtures = jsonContent as Array<FixtureData<EuTaxonomyDataForFinancials>>;
+        testData = getPreparedFixture("company-for-all-types", preparedFixtures);
+      });
+    });
+
+    it("Check", () => {
```
fsoffel, line 32: Test name must be more verbose. Rather do something similar to the text in the `describeIf` function call's first argument.

dataland-frontend/tests/e2e/utils/EuTaxonomyFinancialsUpload.ts
fsoffel: Otherwise cleanup the parameters never used

dataland-frontend/tests/e2e/specs/eu-taxonomy-financials/AddRemoveKpis.ts
```diff
@@ -0,0 +1,113 @@
...
+  cy.get('[data-test="removeSectionButton"]').each(($el, index) => {
+    if (index > 0) {
+      cy.wrap($el).click({ force: true });
+    }
+  });
+
+  cy.get('[data-test="addKpisButton"]').click({ force: true });
+  cy.get('button[data-test="removeSectionButton"]').should("exist").should("have.class", "ml-auto");
+
+  cy.get('[data-test="dataPointToggle"]')
```
fsoffel, line 93: Why do you test the toggling of a datapoint here?

dataland-frontend/src/components/forms/CreateEUTaxonomyForFinancials.vue
```diff
@@ -552,13 +557,39 @@ export default defineComponent({
         this.postEuTaxonomyDataForFinancialsProcessed = false;
         this.messageCount++;
 
+        // JSON.parse/stringify used to clone the formInputsModel in order to stop Proxy refreneces
+        const clonedFormInputsModel = JSON.parse(JSON.stringify(this.formInputsModel)) as ObjectType;
```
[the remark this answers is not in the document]
pioSko2: The problem is that the object has Proxies inside it, which  means that attempts at using shallow cloning like `{ ...obj }` or `Object.assign( {}, obj )` , meant that those Proxies were also inherited (as per nature and intention of a Proxy).
Unfortunately, `structuredClone` cannot be used either since the Proxies inside the object are somehow connected with the Window object, and attempting to do so, throws the following error: `Failed to execute 'structuredClone' on 'Window': #<Object> could not be cloned.`

By using `JSON.parse(JSON.stringify( obj ))` we break the connection to the Proxy instance and receive a clean, new object with just the data needed for the request payload.

It's an older code, but it checks out.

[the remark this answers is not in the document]
fsoffel: All right. Thanks for the explanation :)

dataland-frontend/src/components/forms/CreateEuTaxonomyForFinancials.vue
```diff
@@ -201,70 +201,45 @@
                   </div>
                 </div>
 
-                <div
-                  v-for="financialServiceOption of confirmedSelectedFinancialServiceOptions"
```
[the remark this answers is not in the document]
pioSko2: This was the key issue. Due to Asset Management section being shaped differently, it would cause an error in the form model when being deleted from the kpi sections. That's why (below) I moved the kpi sections into their own scope `name="kpiSectionsModel"` ... (After your comments above, I removed the `v-model=kpiSections"` code)

dataland-frontend/src/components/forms/CreateEuTaxonomyForFinancials.vue
pioSko2: We're running this map only on the kpis found in `kpiSectionsModel` . There's no chance it would be undefined as they are previously created from `euTaxonomyKPIsModel.kpisFieldNameToFinancialServiceType`

dataland-frontend/src/components/forms/CreateEuTaxonomyForFinancials.vue
pioSko2: Changed it to `eligibilityKpi`, but yes, we're getting a single data point here.

dataland-frontend/src/components/forms/CreateEuTaxonomyForFinancials.vue
pioSko2: Hrm. yes.. I think that was my intention. Nice.

dataland-frontend/tests/e2e/specs/eu-taxonomy-financials/AddRemoveKpis.ts
pioSko2: I used an already existing company.

dataland-frontend/src/components/forms/CreateEuTaxonomyForFinancials.vue
```diff
@@ -540,6 +518,34 @@ export default defineComponent({
...
     },
 
+    /**
+     * Parses received eligibility KPIs data
+     * @param receivedFormInputsModelData Received data
+     * @returns ObjectType
+     */
+    extractEligibilityKpis(receivedFormInputsModelData: ObjectType){
+      const fieldNames = Object.values(euTaxonomyKPIsModel.kpisFieldNameToFinancialServiceType);
+      const kpiKeys = Object.keys(euTaxonomyKPIsModel.kpisFieldNameToFinancialServiceType);
```
fsoffel, line 529: these two variables have the same value and never change. We don't need two then.

dataland-frontend/src/components/forms/CreateEuTaxonomyForFinancials.vue
```diff
@@ -540,6 +518,34 @@ export default defineComponent({
...
+      return kpiKeys
+        .map((key: string) => {
+          const fieldName = fieldNames[kpiKeys.indexOf(key)];
```
fsoffel: Wouldnt `euTaxonomyKPIsModel.kpisFieldNameToFinancialServiceType[key]` be simpler than `fieldNames[kpiKeys.indexOf(key)]`?

dataland-frontend/src/components/forms/CreateEuTaxonomyForFinancials.vue
```diff
@@ -570,6 +576,37 @@ export default defineComponent({
       this.onThisPageLinks = [...new Set(this.onThisPageLinksStart.concat(this.selectedFinancialServiceOptions))];
     },
 
+    /**
+     * Converts the kpisSection data to a request-friendly object
+     * @param kpiSectionsModel Kpi Section data
+     * @returns ObjectType
```
fsoffel, line 586: Should also rather describe what you can use it for than the datatype

dataland-frontend/src/components/forms/CreateEuTaxonomyForFinancials.vue
fsoffel: Ok. I find that somewhat hard to notice when you look at the code. Could we emphasize that by putting the generation of the object into a function. Or is this too much overhead?

dataland-frontend/src/components/forms/CreateEuTaxonomyForFinancials.vue
fsoffel: We are not looking on a single data point but on the collection of all eligibility related datapoints of one of the financial service types (`fieldName`).

dataland-frontend/tests/e2e/utils/EuTaxonomyFinancialsUpload.ts
```diff
@@ -265,7 +267,7 @@ export function uploadCompanyViaApiAndEuTaxonomyDataForFinancialsViaForm(
           `/companies/${storedCompany.companyId}/frameworks/${DataTypeEnum.EutaxonomyFinancials}/upload`
         );
         beforeFormFill();
-        fillAndValidateEuTaxonomyForFinancialsUploadForm(testData);
+        formFillSteps(testData);
```
fsoffel, line 270: Now we have three functions we get as parameters in a sequence. We should remove beforeFormFill and afterFormFill and put their content into formFillSteps wherever this function is called.

dataland-frontend/tests/e2e/utils/EuTaxonomyFinancialsUpload.ts
```diff
@@ -245,6 +245,7 @@ export function gotoEditForm(companyId: string, expectIncludedFile: boolean): vo
  * @param companyInformation Company information to be used for the company upload
  * @param testData EU Taxonomy dataset for financial companies to be uploaded
  * @param beforeFormFill is performed before filling the fields of the upload form
+ * @param formFillSteps Steps involved to fill data of the upload form
```
fsoffel, line 248: With this new flexibility of filling the whole form added to the function, we can replace also the corresponding function for the non-financial service data upload, i.e. replace `uploadCompanyViaApiAndEuTaxonomyDataForNonFinancialsViaForm`  by this one

dataland-frontend/src/components/forms/CreateEuTaxonomyForFinancials.vue
```diff
@@ -516,6 +519,37 @@ export default defineComponent({
...
+          const kpis = receivedFormInputsModelData[financialServiceType] as ObjectType;
+          if (kpis) {
+            item[financialServiceType] = {
+              ...item[financialServiceType],
+              ...kpis,
+            };
+          }
+          return item;
+        })
+        .filter((item) => item[Object.keys(item)[0]])
```
fsoffel, line 549: `item[Object.keys(item)[0]]` could be replaced with `Object.values(item)[0]`, right?

dataland-frontend/src/components/forms/CreateEuTaxonomyForFinancials.vue
```diff
@@ -516,6 +519,37 @@ export default defineComponent({
         .catch((e) => console.log(e));
     },
 
+    /**
+     * Parses received financial service types and eligibility KPIs data
+     * @param receivedFormInputsModelData Received data
+     * @returns An object with financial service types and eligibility KPIs which can be directly merged into the formInputsModel, which will then populate expected KPI section fields
```
fsoffel, line 525: `, which will then populate expected KPI section fields` that feels like too much information

dataland-frontend/src/components/forms/CreateEuTaxonomyForFinancials.vue
```diff
@@ -516,6 +519,37 @@ export default defineComponent({
...
+     */
+    extractEligibilityKpis(receivedFormInputsModelData: ObjectType) {
+      const kpiKeys = Object.keys(euTaxonomyKPIsModel.kpisFieldNameToFinancialServiceType);
```
fsoffel, line 528: `kpiKeys` contains the singular of kpi which is not correct. What do you think of `financialServiceTypeKeys`?

dataland-frontend/src/components/forms/CreateEuTaxonomyForFinancials.vue
```diff
@@ -516,6 +519,37 @@ export default defineComponent({
         .catch((e) => console.log(e));
     },
 
+    /**
+     * Parses received financial service types and eligibility KPIs data
+     * @param receivedFormInputsModelData Received data
+     * @returns An object with financial service types and eligibility KPIs which can be directly merged into the formInputsModel, which will then populate expected KPI section fields
+     */
+    extractEligibilityKpis(receivedFormInputsModelData: ObjectType) {
```
fsoffel, line 527: We are not only extracting eligibility kpis but also other financial service specific kpis. I'd be fine with just dropping 'eligibility' from the method's name.

dataland-frontend/src/components/forms/CreateEuTaxonomyForFinancials.vue
```diff
@@ -516,6 +519,37 @@ export default defineComponent({
...
+      const kpiKeys = Object.keys(euTaxonomyKPIsModel.kpisFieldNameToFinancialServiceType);
+      return kpiKeys
+        .map((key: string) => {
```
fsoffel, line 530: `financialServiceTypeKey` for consistency then, maybe? Or what is the benefit of shorter names?

dataland-frontend/src/components/forms/CreateEuTaxonomyForFinancials.vue
```diff
@@ -201,81 +201,78 @@
...
-                      :name="euTaxonomyKPIsModel?.kpisFieldNameToFinancialServiceType[financialServiceOption.value]"
-                      type="group"
-                    >
-                      <div
-                        v-for="kpiTypeEligibility of euTaxonomyKPIsModel.eligibilityKpis"
-                        :key="kpiTypeEligibility"
-                        :data-test="kpiTypeEligibility"
-                        class="uploadFormSection"
+                      <FormKit
+                        :name="euTaxonomyKPIsModel?.kpisFieldNameToFinancialServiceType[financialServiceOption.value]"
```
fsoffel, line 250: We would save us some complexity if the name of this formkit group would just be eligibilityKpis

fsoffel: Oh, it doesn't. But it would make things clearer IMO.

dataland-frontend/src/components/forms/CreateEuTaxonomyForFinancials.vue
```diff
@@ -546,6 +580,42 @@ export default defineComponent({
       this.onThisPageLinks = [...new Set(this.onThisPageLinksStart.concat(this.selectedFinancialServiceOptions))];
     },
 
+    /**
+     * Converts the kpisSection data to a request-friendly object
+     * @param kpiSections Kpi Section data
+     * @returns ObjectType
+     */
+    makeEligibilityKpis(kpiSections: ObjectType) {
```
fsoffel, line 588: Again, its not all about the eligibility kpis. I propose `buildKpis`.

fsoffel: Or `convert...`

dataland-frontend/src/components/forms/CreateEuTaxonomyForFinancials.vue
```diff
@@ -546,6 +580,42 @@ export default defineComponent({
...
+      const eligibilityKpis = Object.keys(kpiSections)
+        .map((key) => {
```
fsoffel, line 590: financialServiceTypeKey > key

dataland-frontend/src/components/forms/CreateEuTaxonomyForFinancials.vue
```diff
@@ -546,6 +580,42 @@ export default defineComponent({
...
+          const eligibilityKpi = {
```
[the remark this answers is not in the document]
fsoffel: Why would you store the key here again? Does not fit the name anyways.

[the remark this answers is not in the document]
fsoffel: Why even make an object here anyways?

dataland-frontend/src/components/forms/CreateEuTaxonomyForFinancials.vue
```diff
@@ -546,6 +580,42 @@ export default defineComponent({
...
+            financialServiceType: euTaxonomyKPIsModel.kpisFieldNameToFinancialServiceType[key] as string,
+          };
+          const section = kpiSections[eligibilityKpi.key] as ObjectType;
+          const field = section[eligibilityKpi.financialServiceType];
+          return { [eligibilityKpi.financialServiceType]: field };
+        })
+        .reduce((all, one) => ({ ...all, ...one }));
+
+      const kpis = Object.keys(kpiSections)
+        .filter((key) => key !== "assetManagementKpis")
```
fsoffel, line 602: financialServiceTypeKey > key

dataland-frontend/src/components/forms/CreateEuTaxonomyForFinancials.vue
```diff
@@ -546,6 +580,42 @@ export default defineComponent({
...
+        .map((key) => {
+          const kpi = { [key]: kpiSections[key] };
+          const kpiKey = Object.keys(kpi)[0];
```
fsoffel, line 605: kpiKey = key? Seems redundant.

fsoffel: Aaah, if there is a kpiSection for it. Otherwise undefined. I would check explicitely if that section exists and only use the key variable for readability.

dataland-frontend/src/components/forms/CreateEuTaxonomyForFinancials.vue
```diff
@@ -546,6 +580,42 @@ export default defineComponent({
...
+          if (kpiKey) {
+            const financialServiceType = (euTaxonomyKPIsModel.kpisFieldNameToFinancialServiceType as ObjectType)[
+              kpiKey
+            ];
+            delete (kpi[kpiKey] as ObjectType)[financialServiceType as string];
```
fsoffel, line 610: I don't get it. There is nothing to delete, I think.

dataland-frontend/tests/e2e/specs/eu-taxonomy-financials/AddRemoveKpis.ts
```diff
@@ -88,7 +88,6 @@ function fillAndValidateEuTaxonomyForFinancialsUploadForm(data: EuTaxonomyDataFo
     }
   });
 
-  cy.get('[data-test="addKpisButton"]').click({ force: true });
```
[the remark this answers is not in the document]
pioSko2: This was a duplicate line.  I've now commented the test to highlight which sections of the code do what actions

dataland-frontend/src/components/forms/CreateEuTaxonomyForFinancials.vue
pioSko2: Yes hehe Done.

dataland-frontend/src/components/forms/CreateEuTaxonomyForFinancials.vue
pioSko2: Code readability only. I'll change it for consistency.

pioSko2: Actually, I think it will be better to leave it as `key`. We already know it is coming from `financialServiceTypeKeys`

dataland-frontend/src/components/forms/CreateEuTaxonomyForFinancials.vue
```diff
@@ -516,6 +519,37 @@ export default defineComponent({
...
+     * Parses received financial service types and eligibility KPIs data
+     * @param receivedFormInputsModelData Received data
+     * @returns An object with financial service types and eligibility KPIs which can be directly merged into the formInputsModel, which will then populate expected KPI section fields
+     */
+    extractEligibilityKpis(receivedFormInputsModelData: ObjectType) {
+      const kpiKeys = Object.keys(euTaxonomyKPIsModel.kpisFieldNameToFinancialServiceType);
+      return kpiKeys
+        .map((key: string) => {
+          const financialServiceType = Object.values(euTaxonomyKPIsModel.kpisFieldNameToFinancialServiceType)[
+            kpiKeys.indexOf(key)
```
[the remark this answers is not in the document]
pioSko2: That's kind of the point of it. We don't know what the order is and in this case we don't need to. It just gets the needed index, whatever it is.

dataland-frontend/src/components/forms/CreateEuTaxonomyForFinancials.vue
fsoffel: But back then `elibilityKpis` existed multiple times as one object. Here it would be different subobjects

dataland-frontend/src/components/forms/CreateEuTaxonomyForFinancials.vue
pioSko2: Before the delete we'll have an object like this:
```
{
    "creditInstitutionKpis": {
        "CreditInstitution": { ... },
        "tradingPortfolio": { ... },
        "interbankLoans": { ... },
        "tradingPortfolioAndInterbankLoans": { ... },
        "greenAssetRatio": { ... }
    }
}
````
... and we need to delete `CreditInstitution` in order for the request object to be valid.

File: dataland-frontend/src/components/forms/CreateEuTaxonomyForFinancials.vue (hunk 8 above)
Comment (pioSko2) [reply dataland-frontend/src/components/forms/CreateEuTaxonomyForFinancials.vue]: Done.

File: dataland-frontend/src/components/forms/CreateEuTaxonomyForFinancials.vue (hunk 23 above)
Comment (pioSko2) [reply dataland-frontend/src/components/forms/CreateEuTaxonomyForFinancials.vue]: Done.

File: dataland-frontend/src/components/forms/CreateEuTaxonomyForFinancials.vue (hunk 24 above)
Comment (pioSko2) [reply dataland-frontend/src/components/forms/CreateEuTaxonomyForFinancials.vue]: Done.

File: dataland-frontend/tests/e2e/utils/EuTaxonomyFinancialsUpload.ts (hunk 26 above)
Comment (pioSko2) [reply dataland-frontend/tests/e2e/utils/EuTaxonomyFinancialsUpload.ts:248]: Done.

File: dataland-frontend/src/components/forms/CreateEuTaxonomyForFinancials.vue (hunk 28 above)
Comment (pioSko2) [reply dataland-frontend/src/components/forms/CreateEuTaxonomyForFinancials.vue]: Done.

File: dataland-frontend/src/components/forms/CreateEuTaxonomyForFinancials.vue (hunk 29 above)
Comment (pioSko2) [reply dataland-frontend/src/components/forms/CreateEuTaxonomyForFinancials.vue]: Done.

File: dataland-frontend/src/components/forms/CreateEuTaxonomyForFinancials.vue (hunk 30 above)
Comment (pioSko2) [reply dataland-frontend/src/components/forms/CreateEuTaxonomyForFinancials.vue]: Done.

File: dataland-frontend/src/components/forms/CreateEuTaxonomyForFinancials.vue (hunk 33 above)
Comment (pioSko2) [reply dataland-frontend/src/components/forms/CreateEuTaxonomyForFinancials.vue]: Done.

File: dataland-frontend/src/components/forms/CreateEuTaxonomyForFinancials.vue (hunk 34 above)
Comment (pioSko2) [reply dataland-frontend/src/components/forms/CreateEuTaxonomyForFinancials.vue]: Done.

File: dataland-frontend/src/components/forms/CreateEuTaxonomyForFinancials.vue (hunk 35 above)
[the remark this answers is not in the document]
Comment (pioSko2) [reply dataland-frontend/src/components/forms/CreateEuTaxonomyForFinancials.vue]: Done.

File: dataland-frontend/src/components/forms/CreateEuTaxonomyForFinancials.vue (hunk 36 above)
Comment (pioSko2) [reply dataland-frontend/src/components/forms/CreateEuTaxonomyForFinancials.vue]: Done.

File: dataland-frontend/src/components/forms/CreateEuTaxonomyForFinancials.vue (hunk 37 above)
Comment (pioSko2) [reply dataland-frontend/src/components/forms/CreateEuTaxonomyForFinancials.vue]: Done.

File: dataland-frontend/tests/e2e/specs/eu-taxonomy-financials/AddRemoveKpis.ts (hunk 19 above)
Comment (pioSko2) [reply dataland-frontend/tests/e2e/specs/eu-taxonomy-financials/AddRemoveKpis.ts:94]: Removed. Done.

File: dataland-frontend/tests/e2e/utils/EuTaxonomyFinancialsUpload.ts (hunk 25 above)
Comment (pioSko2) [reply dataland-frontend/tests/e2e/utils/EuTaxonomyFinancialsUpload.ts:270]: Done.

File: dataland-frontend/src/components/forms/CreateEuTaxonomyForFinancials.vue (hunk 33 above)
Comment (fsoffel) [reply dataland-frontend/src/components/forms/CreateEuTaxonomyForFinancials.vue]: There is still the word `eligibility` inside the name of the function

File: dataland-frontend/tests/e2e/utils/EuTaxonomyUpload.ts
Comment (fsoffel) [inline dataland-frontend/tests/e2e/utils/EuTaxonomyUpload.ts:1]: I think this file can be removed :)

File: dataland-frontend/src/components/forms/CreateEuTaxonomyForFinancials.vue (hunk 38 above)
Comment (fsoffel) [reply dataland-frontend/src/components/forms/CreateEuTaxonomyForFinancials.vue]: Got it, thanks.

File: dataland-frontend/src/components/forms/CreateEuTaxonomyForFinancials.vue (hunk 40 above)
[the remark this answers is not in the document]
Comment (fsoffel) [reply dataland-frontend/src/components/forms/CreateEuTaxonomyForFinancials.vue]: But could it replaced with the `ObjectType` workaround? It would be more consistent then.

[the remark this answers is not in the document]
Comment (pioSko2) [reply dataland-frontend/src/components/forms/CreateEuTaxonomyForFinancials.vue]: Yes I agree that it should work. However 2 things. When I changed it to use ObejctType, then in the next line eslint complained that there is an "Unsafe assignment of an `any` value."
Secondly, I strongy believe we should move away from the ObjectType workaround and start using strict types/interfaces as much as possible. What's the point of having TypeScript with type checking if we're cheating every second line to stop lint warnings :)

I even tried something like this, but still kept getting eslint warnings in the last line:
```
const financialServiceType = (euTaxonomyKPIsModel.kpisFieldNameToFinancialServiceType as ObjectType)[key];
const receivedEligibilityKpis = receivedFormInputsModelData.eligibilityKpis as ObjectType;
const kpis = (receivedEligibilityKpis as ObjectType)[financialServiceType];
```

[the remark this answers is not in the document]
fsoffel: That sounds reasonable :( Then it is fine as it is. I'll look out for other solutions.

PR #257, opened 2023-07-12
Dala 1849 introduce flyway

fsoffel:
# Pull Request \<Title>
`<Description here>`


dataland-backend/src/main/kotlin/db/migration/V2__GenerateCompanyNameIdentifierIndex.kt
```diff
@@ -0,0 +1,31 @@
...
+
+/**
+ * This migration script creates an index for a faster search for company names and identifiers
+ */
+@Suppress("ClassNaming")
+class V2__GenerateCompanyNameIdentifierIndex : BaseJavaMigration() {
+    override fun migrate(context: Context?) {
+        context!!.connection.createStatement().execute(
+            "CREATE EXTENSION IF NOT EXISTS pg_trgm; " +
+                "CREATE INDEX IF NOT EXISTS idx_company_identifier " +
```
ahoecherl, line 14: As discussed the Flyway way of doing this would be without the _IF NOT EXISTS_

dataland-frontend/tests/e2e/specs/framework-view-and-upload-journey/UploadFrameworkDataJourney.ts
```diff
@@ -21,6 +21,7 @@ describe("As a user, I expect the dataset upload process to behave as I expect",
     "",
     {
       executionEnvironments: ["developmentLocal", "ci", "developmentCd"],
+      onlyExecuteOnReset: false,
```
ahoecherl, line 24: "OnReset" is a bit ambiguous I'd prefer "OnDatabaseWipe" or similar

dataland-frontend/tests/e2e/support/TestUtility.ts
```diff
@@ -2,6 +2,7 @@ import { Suite } from "mocha";
 
 export interface ExecutionConfig {
   executionEnvironments: Array<ExecutionEnvironment>;
+  onlyExecuteOnReset: boolean;
```
ahoecherl, line 5: Some way to have a default value of false here?

dataland-internal-storage/src/main/kotlin/db/migration/V2__MigrateEuTaxonomyNonFinancialsWithAbsoluteValues.kt
```diff
@@ -0,0 +1,77 @@
...
+import org.dataland.datalandbackend.openApiClient.model.DataTypeEnum
+import org.flywaydb.core.api.migration.BaseJavaMigration
+import org.flywaydb.core.api.migration.Context
+import org.json.JSONObject
+
+/**
+ * This migration script updates the EU Taxonomy for non-financials data model switching
+ * from a field called value for percentages only to a structure holding the absolute value of a cash flow type as well
+ */
+@Suppress("ClassNaming")
```
ahoecherl, line 13: Any possibility to also change the linter settings to not make this Suppress statement necessary?

dataland-internal-storage/src/test/kotlin/org/dataland/datalandinternalstorage/migrations/V2__MigrationTest.kt
```diff
@@ -0,0 +1,101 @@
...
+import org.junit.jupiter.api.Test
+import org.mockito.ArgumentMatchers.any
+import org.mockito.Mockito.mock
+import org.mockito.Mockito.`when`
+import java.sql.Connection
+import java.sql.ResultSet
+import java.sql.Statement
+
+@Suppress("ClassNaming")
+class V2__MigrationTest {
```
ahoecherl, line 17: Shouldn't the class name be V2__MigrateEuTaxonomyNonFinancialsWithAbsoluteValuesTest

docker-compose.yml
```diff
@@ -146,6 +146,7 @@ services:
       PROXY_PRIMARY_URL: ${PROXY_PRIMARY_URL:-local-dev.dataland.com}
       INVITATION_REQUEST_RECEIVERS: ${INVITATION_REQUEST_RECEIVERS}
       INVITATION_REQUEST_CC: ${INVITATION_REQUEST_CC}
+      BACKEND_MIGRATION_BASELINE_VERSION: ${BACKEND_MIGRATION_BASELINE_VERSION:-1}
```
ahoecherl, line 149: The :-1 sets a default value of 1? Does that mean that we no longer need the default environment variable? If we still need it than I guess that the :-1 doesn't actually have the desired effect and is not needed.

fsoffel: We use the :-1 for the local environment, I don't want to enforce everyone to set this env.

fsoffel: On dev PCs

dataland-internal-storage/src/test/kotlin/org/dataland/datalandinternalstorage/migrations/V2__MigrationTest.kt
fsoffel: It doesn't matter too much since we get the uniqueness from the version number. But if you like, I can rename it.

dataland-internal-storage/src/main/kotlin/db/migration/V2__MigrateEuTaxonomyNonFinancialsWithAbsoluteValues.kt
fsoffel: This should be possible. I'll look into it.

fsoffel: Das war ja einfach. Da war sogar schon eine RegEx angelegt.

dataland-frontend/tests/e2e/support/TestUtility.ts
fsoffel: We could make this parameter optional I guess. However, this is intentional. My reasoning is that when using describeIf the user should be aware of all the options.

dataland-frontend/tests/e2e/specs/framework-view-and-upload-journey/UploadFrameworkDataJourney.ts
fsoffel: Agree on the name, changed it to `...OnDatabaseReset`, wipe sounds informal IMO. For the rest I refer to my other answer.

dataland-backend/src/main/kotlin/db/migration/V2__GenerateCompanyNameIdentifierIndex.kt
fsoffel: Aaah right. On it.

fsoffel: Changed it. However I want to test if this works. Also we should set the baseline version on preview to 2 now. Since it has an index already. But I'll check.

ahoecherl (approved):

PR #293, opened 2023-09-07
Dala 2132 wrong values displayed on p2p view page

SiegfriedSobkowiak:
# Pull Request \<Title>
`<Description here>`
- [ ] IF there are changes done to the framework data models or to a database entity class, the following steps were completed
  - [ ] The main branch is deployed to a dev server with `Reset non-user related Docker Volumes & Re-populate` turned on
  - [ ] Immediately after that the feature branch is deployed to the same server with `Reset non-user related Docker Volumes & Re-populate` turned off
are merged


dataland-frontend/tests/component/components/resources/frameworkDataSearch/ThreeLayerDataTable.cy.ts
```diff
@@ -218,4 +227,63 @@ describe("Component test for the NewEUTaxonomy Page", () => {
...
+          energyMix: 22,
+        },
+        technology: {},
+        material: {},
+      },
+    };
+
+    cy.mountWithPlugins(ThreeLayerDataTable, {
+      keycloak: minimalKeycloakMock({}),
+      // eslint-disable-next-line @typescript-eslint/ban-ts-comment
```
SiegfriedSobkowiak, line 270: Removing this disable and running the linter yields no errors locally. Why is it required?

dataland-frontend/src/components/resources/frameworkDataSearch/TwoLayerDataTable.vue
```diff
@@ -48,71 +48,75 @@
...
-              "
-            >
-              {{ slotProps.data.content[reportingPeriodWithDataId.dataId] }} %</span
-            >
             <span
-              v-else-if="
-                typeof slotProps.data.content[reportingPeriodWithDataId.dataId] === 'object' &&
-                slotProps.data.content[reportingPeriodWithDataId.dataId]?.value !== undefined
-              "
+              :data-test="`${reportingPeriodWithDataId.reportingPeriod}_${slotProps.data.categoryKey}_${slotProps.data.kpiKey}`"
```
SiegfriedSobkowiak, line 52: I can confirm that the error described in the ticket is gone. However, I am not sure how this change fixes it. If I see it correctly there was only a span added that contains only data-test information. How does this lead to resolving the bug?

dataland-frontend/tests/component/components/resources/frameworkDataSearch/ThreeLayerDataTable.cy.ts
pioSko2: I get errors locally:

image

image

dataland-frontend/src/components/resources/frameworkDataSearch/TwoLayerDataTable.vue
pioSko2: It is only needed for the test. Adding a span in this spot allows us to wrap every value created within the conditionals nested within. Otherwise we'd need to wrap each value with their own span within each conditional.
I added this span to improve  targeting of specific values within tests. It was possible before, yes, but rather difficult and cumbersome code. This way we can directly target each cell easily.

dataland-frontend/tests/component/components/resources/frameworkDataSearch/ThreeLayerDataTable.cy.ts
SiegfriedSobkowiak: Sorry I was not clear in my comment. I meant removing both comment lines. The eslint disable and the ts ignore. When I do that locally the linter is still fine and the component tests are still green.

pioSko2: Removing both I get an even chunkier message :)
image

Marc mentioned this during his overview of component testing that these comments are necessary to shut typescript up.

SiegfriedSobkowiak: I am wondering why I do not get any errors locally. But for this PR it does not matter.

PR #305, opened 2023-09-22
Dala 2091 allow special characters in pdf names

ahoecherl:
# Pull Request \ Dala 2091 allow special characters in pdf names
Allows documents with with a wider range of file names to be uploaded. Basically any name that is valid under Windows.
- [ ] IF there are changes done to the framework data models or to a database entity class, the following steps were completed
  - [ ] The main branch is deployed to a dev server with `Reset non-user related Docker Volumes & Re-populate` turned on
  - [ ] Immediately after that the feature branch is deployed to the same server with `Reset non-user related Docker Volumes & Re-populate` turned off
are merged


.githooks/pre-commit
fsoffel, line 1: Why is this empty?

dataland-frontend/src/components/general/InvalidFileSelectionDialog.vue
```diff
@@ -0,0 +1,48 @@
...
+
+  methods: {
+    /**
+     * Gets all the data that is passed down by the component which has opened this modal and stores it in the
+     * component-level data object.
+     */
+    getDataFromParentAndSet() {
+      const dialogRefToDisplay = this.dialogRef as DynamicDialogInstance;
+      const dialogRefData = dialogRefToDisplay.data as {
+        message: string;
```
fsoffel, line 39: `message` seems unused.

dataland-frontend/src/components/forms/parts/UploadReports.vue
```diff
@@ -147,42 +160,60 @@ export default defineComponent({
       }
     },
     /**
-     * Handles selection of a file by the user. If duplicates are found in the selection, this is handled.
+     * Handles selection of files by the user. If invalid file names are found in the selection, this is handled.
```
fsoffel, line 163: Duplicates are also handled.

dataland-frontend/src/components/forms/parts/UploadReports.vue
```diff
@@ -147,42 +160,60 @@ export default defineComponent({
...
+            invalidityReason: FileNameInvalidityReason.Duplicate,
+          });
+        } else {
+          existingFileNamesCollector.add(fileName);
         }
+      }
 
-        this.handleReportDuplicates([...duplicatesWithIndex].reverse());
+      if (invalidFileNamesWithIndexAndReason.length > 0) {
+        this.handleInvalidFileNames([...invalidFileNamesWithIndexAndReason].reverse());
```
fsoffel, line 194: The name `handleInvalidFiles` would fit. This one does not.

dataland-frontend/src/components/forms/parts/UploadReports.vue
```diff
@@ -147,42 +160,60 @@ export default defineComponent({
...
      */
-    handleUpdatedDocumentsSelectedForUpload(documents: DocumentToUpload[]) {
-      this.documentsToUpload = documents;
-      const duplicatesWithIndex: DuplicateWithIndex[] = [];
+    handleUpdatedDocumentsSelectedForUpload(selectedDocumentsForUpload: DocumentToUpload[]) {
+      this.documentsToUpload = selectedDocumentsForUpload;
 
-      if (this.areDuplicatesAmongReferenceableReportNames()) {
-        const foundExistingRecords = new Set<string>();
+      const invalidFileNamesWithIndexAndReason: InvalidFileNameWithIndexAndReason[] = [];
```
fsoffel, line 170: `invalidFileNamesWithIndexAndReason` is a bit misleading that this might only contain files with invalid filenames. What about `nameIndexAndReasonOfInvalidFiles` or `invalidFilesByNameIndexAndReason`?

dataland-frontend/src/utils/Constants.ts
```diff
@@ -30,4 +30,7 @@ export const TIME_DELAY_BETWEEN_SUBMIT_AND_NEXT_ACTION_IN_MS = 2000;
 
 export const UPLOAD_FILE_SIZE_DISPLAY_DECIMALS = 2;
 
+// Same as standard for windows machines. File name can't have one of the special characters <>:"|?/\ and shouldn't end with a whitespace or .
```
fsoffel, line 33: Shouldn't this be the minimum from what is accepted on any major OS? windows, macos, ubuntu/linux

dataland-frontend/src/utils/Constants.ts
```diff
@@ -30,4 +30,7 @@ export const TIME_DELAY_BETWEEN_SUBMIT_AND_NEXT_ACTION_IN_MS = 2000;
...
+export const REGEX_FOR_FILE_NAMES = /^(?=.{1,254}$)[^<>:"|?/*\\]+[^<>:"|?/*\\.\s]$/;
```
fsoffel, line 34: This should be used in the backend as well

dataland-frontend/src/utils/Constants.ts
fsoffel: Add `*` to the list in the comment

dataland-frontend/src/utils/Constants.ts
fsoffel, line 34: Whitespace in the beginning of a filename seems not to be allowed as well.

dataland-frontend/src/components/general/InvalidFileSelectionDialog.vue
emanueldfine: Removed it.

dataland-frontend/src/utils/Constants.ts
fsoffel: Apparently, MacOs and Linux only disallow very few characters in filenames which are already included in the ones forbidden by windows.

.githooks/pre-commit
ahoecherl: I changed the file with chmod to be executable on a linux system and for some reason, git was of the opinion that that was a change that required a commit. Maybe the bits of the file actually changed but the content indeed did not.

fsoffel: All right. As long as the script is still executable as the regular githook, this is fine.

dataland-frontend/src/components/forms/parts/UploadReports.vue
emanueldfine: I have extended this js doc to make clear that an invalid file name can have two reasons (forbidden character or duplicate name)

dataland-frontend/src/components/forms/parts/UploadReports.vue
emanueldfine: How about `handleFilesWithInvalidNames()` ?
Because the files themselves (meaning their content) is not detected as invalid here. It is just the names that are not ok.

dataland-frontend/src/components/forms/parts/UploadReports.vue
emanueldfine: I renamed the variables and the type to "NameIndexAndReasonOfInvalidFile(s)"

dataland-frontend/src/components/forms/parts/UploadReports.vue
fsoffel: Perfect. I guess I misunderstood the meaning of invalid file names first.

dataland-frontend/src/components/forms/parts/UploadReports.vue
fsoffel: 👍 fair

dataland-frontend/src/components/forms/parts/UploadReports.vue
fsoffel: 👍

dataland-document-manager/src/main/kotlin/org/dataland/documentmanager/services/PdfVerificationService.kt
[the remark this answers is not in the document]
emanueldfine: I have adjusted this.  Now it is a "file verification" service and all the js docs speak of "files" until it is asserted that it is indeed a pdf.

dataland-frontend/src/utils/Constants.ts
emanueldfine: I added the `*` to the comment.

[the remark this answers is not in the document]
emanueldfine: `*` has been added

dataland-document-manager/src/main/kotlin/org/dataland/documentmanager/services/FileVerificationService.kt
```diff
@@ -8,40 +8,40 @@ import org.springframework.web.multipart.MultipartFile
...
-     * A function that performs surface-level checks to ensure that an uploaded document is indeed a PDF.
+     * A function that performs surface-level checks to ensure that an uploaded file is indeed a PDF.
      * This function does not enforce anything else.
      * In particular a file passing this function is in no way guaranteed to be safe to open.
      */
-    fun assertThatDocumentLooksLikeAPdf(document: MultipartFile, correlationId: String) {
+    fun assertThatFileLooksLikeAPdf(file: MultipartFile, correlationId: String) {
         try {
-            checkIfPdfDocumentIsEmpty(document.bytes, correlationId)
+            checkIfPotentialPdfFileIsEmpty(file.bytes, correlationId)
```
fsoffel, line 24: I would completely detach anything here from the pdf format. We can upload anything right now. So why not just `checkIfFileIsEmpty`.

dataland-frontend/src/utils/Constants.ts
emanueldfine: Are you sure? They way I understand the regex, a whitespace at the beginning is not excluded and therefore it should be allowed.
If you have doubts, let's talk about this.

emanueldfine: Btw I have simplified the regex (without changing its outcome) to have it a little bit shorter.

fsoffel: When I try to add whitespace to a filename in windows after the last proper character, it gets automatically removed. When I do the same at the front the same happens. And yes, the regex does not reflect that.

dataland-frontend/src/components/forms/parts/UploadReports.vue
emanueldfine: Renamed

dataland-document-manager/src/main/kotlin/org/dataland/documentmanager/services/FileVerificationService.kt
emanueldfine: Because the emptiness-check only works, if the file can be interpreted as pdf.
But because the file can also be something else, it checks if the "potential" pdf is empty.

dataland-document-manager/src/main/kotlin/org/dataland/documentmanager/services/FileVerificationService.kt
```diff
@@ -8,40 +8,40 @@ import org.springframework.web.multipart.MultipartFile
...
     private val logger = LoggerFactory.getLogger(javaClass)
 
     /**
-     * A function that performs surface-level checks to ensure that an uploaded document is indeed a PDF.
+     * A function that performs surface-level checks to ensure that an uploaded file is indeed a PDF.
      * This function does not enforce anything else.
      * In particular a file passing this function is in no way guaranteed to be safe to open.
      */
-    fun assertThatDocumentLooksLikeAPdf(document: MultipartFile, correlationId: String) {
+    fun assertThatFileLooksLikeAPdf(file: MultipartFile, correlationId: String) {
```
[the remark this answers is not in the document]
emanueldfine: It seems like that at first sight.
But in reality, it does two checks:
1) It checks (tries) if it can be parsed as pdf
2) If it can be parsed as pdf, it is checked if its is empty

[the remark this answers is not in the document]
emanueldfine: Maybe name it "assertThatFileIsParsableAsPdfWithAtLeastOnePage" ?

dataland-frontend/src/utils/Constants.ts
emanueldfine: Ahhh ok your first comment meant that it is not allowed as well on WINDOWS.
I thought you mean in the regex.

But then I should explicitly forbid a whitespace at the beginning via the Regex, right?

dataland-document-manager/src/main/kotlin/org/dataland/documentmanager/services/FileVerificationService.kt
fsoffel: Aaaaah. I see. It is not just a check for the file size.

dataland-document-manager/src/main/kotlin/org/dataland/documentmanager/services/FileVerificationService.kt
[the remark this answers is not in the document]
fsoffel: I'm fine then as is. 👍

dataland-document-manager/src/main/kotlin/org/dataland/documentmanager/services/FileVerificationService.kt
```diff
@@ -0,0 +1,49 @@
...
+import org.slf4j.LoggerFactory
+import org.springframework.stereotype.Component
+import org.springframework.web.multipart.MultipartFile
+import java.io.IOException
+
+/**
+ * A service for performing basic sanity checks on files uploaded by user
+ */
+@Component
+class FileVerificationService {
```
fsoffel, line 14: Should we revert that renaming commit then?

dataland-frontend/src/utils/Constants.ts
fsoffel: Yup

emanueldfine: Ok pls check my last commit.  It should forbid a whitespace at the beginning now and also mention that in the comment.

dataland-document-manager/src/main/kotlin/org/dataland/documentmanager/services/FileVerificationService.kt
emanueldfine: However you like.   I am also ok with the current name, because before the verificator does its job, you are not 100% sure that you have an actual pdf at hand.   So at this point, it is indeed a file verificator.

On the other hand, it is looking for pdfs, so pdf verificator would also be ok (if the name shall contain the thing it is looking for, rather than the thing it is actually scanning).

So you tell me.  I am 50 50.

dataland-frontend/src/utils/Constants.ts
emanueldfine: From a consistency point of view: Yes.
But I think it won't break anything if we don't have it in the backend.

Or do you see potential problems that could occur?

fsoffel: That does forbid whitespaces at any position. What about `^[^<>:"|?/*\\.\s][^<>:"|?/*\\]{0,253}[^<>:"|?/*\\.\s]$`?
This would however enforce a filename with at least 2 symbols which is probably fine.

dataland-document-manager/src/main/kotlin/org/dataland/documentmanager/services/FileVerificationService.kt
fsoffel: Sorry for the effort done. But I'd prefer the original version :/

dataland-frontend/src/utils/Constants.ts
emanueldfine: That forbids a `.` at the beginning, and allows a total fo 255 characters.

Check out my last commit.  Is there any problem you see in the regex?

emanueldfine: For got the "."  at the end.  Another commit

dataland-document-manager/src/main/kotlin/org/dataland/documentmanager/services/FileVerificationService.kt
emanueldfine: Ok no problem. Will start doing it now

dataland-frontend/src/utils/Constants.ts
emanueldfine: Ah ok yes that would indeed be a problem.  Good point.  WIll include it in the backend

fsoffel: The `.` was an accident.

fsoffel: I reduced the center block to max 253 symbols

dataland-document-manager/src/main/kotlin/org/dataland/documentmanager/services/FileVerificationService.kt
emanueldfine: Done

PR #357, opened 2023-11-13
Dala 2632 gleif isin extension

ahoecherl:
# Pull Request \<Title>
`<Description here>`
  - [ ] The version of main currently active on prod is deployed to a dev server with `Reset non-user related Docker Volumes & Re-populate` turned on
  - [ ] The data from prod is migrated using `./migrateData.sh dataland.com <SOURCE_API_KEY> <TARGET> <TARGET_API_KEY>` 
  - [ ] The feature branch is deployed to the same server with `Reset non-user related Docker Volumes & Re-populate` turned off
are merged


.githooks/pre-commit
fsoffel, line 1: What changed in this file? I did not find any change in permissions.

dataland-batch-manager/src/main/resources/application.properties
```diff
@@ -2,5 +2,7 @@ dataland.dataland-batch-manager.client-id=dataland-batch-manager
 dataland.dataland-batch-manager.client-secret=${DATALAND_BATCH_MANAGER_CLIENT_SECRET}
 dataland.dataland-batch-managet.get-all-gleif-companies.flag-file=${GET_ALL_GLEIF_COMPANIES_FLAG_FILE}
 dataland.backend.base-url=${INTERNAL_BACKEND_URL}
+dataland.dataland-batch-manager.mapping-file=/deltafiles/isinMapping.csv
 dataland.keycloak.base-url=http://keycloak:8080/keycloak
 gleif.download.baseurl=https://goldencopy.gleif.org/api/v2/golden-copies/publishes/lei2
+gleif.mapping.download.baseurl=https://mapping.gleif.org/api/v2/isin-lei
```
fsoffel, line 8: 1.) Since this is (in contrast to `gleif.download.baseurl`) already the complete url it should not say `baseurl` but rather `url` only.

2.) Maybe `isinmapping` instead of `mapping`?

dataland-batch-manager/src/main/resources/application.properties
```diff
@@ -2,5 +2,7 @@ dataland.dataland-batch-manager.client-id=dataland-batch-manager
 dataland.dataland-batch-manager.client-secret=${DATALAND_BATCH_MANAGER_CLIENT_SECRET}
 dataland.dataland-batch-managet.get-all-gleif-companies.flag-file=${GET_ALL_GLEIF_COMPANIES_FLAG_FILE}
 dataland.backend.base-url=${INTERNAL_BACKEND_URL}
+dataland.dataland-batch-manager.mapping-file=/deltafiles/isinMapping.csv
```
fsoffel, line 5: isinmapping-file?

dataland-batch-manager/src/main/resources/application-development.properties
```diff
@@ -2,5 +2,6 @@ dataland.dataland-batch-manager.client-id=dataland-batch-manager
 dataland.dataland-batch-manager.client-secret=password
 dataland.dataland-batch-managet.get-all-gleif-companies.force=false
 dataland.dataland-batch-managet.get-all-gleif-companies.flag-file=./this-file-does-not-exist
+dataland.dataland-batch-manager.mapping-file=./deltafiles/isinMapping.csv
```
fsoffel, line 5: This should be an absolute path as well.

docker-compose.yml
```diff
@@ -412,6 +412,7 @@ services:
       DATALAND_BATCH_MANAGER_CLIENT_SECRET: ${DATALAND_BATCH_MANAGER_CLIENT_SECRET}
     volumes:
       - ./dataland-batch-manager/config:/config
+      - batch_manager_data_volume:/deltafiles
```
fsoffel, line 415: TODO for me: test delta file persistance

.githooks/pre-commit
fsoffel: Did someone run some chmod stuff? why is this reflected here.

dataland-batch-manager/src/main/kotlin/org/dataland/datalandbatchmanager/service/GleifGoldenCopyIngestor.kt
```diff
@@ -94,6 +107,31 @@ class GleifGoldenCopyIngestor(
         logger.info("Finished processing of file $csvFile in ${getExecutionTime(start)}.")
     }
 
+    private fun prepareMappingFile() {
+        logger.info("Starting LEI-ISIN mapping update cycle for latest file.")
+        val tempFile = File.createTempFile("gleif_mapping_update", ".csv")
+        processMappingFile(tempFile, gleifApiAccessor::getIsinMappingFile)
+    }
+
+    @Synchronized
```
fsoffel, line 116: @SiegfriedSobkowiak do I understand correctly that we use the `@Synchronized` annotation here (and above for the gleif diff) to prevent this from running more than once at the same time, e.g. if last sundays task is not finished yet, do not start the next update yet, as an extreme case?

dataland-batch-manager/src/main/kotlin/org/dataland/datalandbatchmanager/service/GleifApiAccessor.kt
```diff
@@ -35,6 +37,17 @@ class GleifApiAccessor(
         downloadFileFromGleif("latest.csv", targetFile, "full Golden Copy File")
     }
 
+    /**
+     * Downloads the latest complete Lei-ISIN mapping file
+     * @param targetFile the local target file to be written
+     */
+    fun getIsinMappingFile(targetFile: File) {
+        val apiText = URL(gleifMappingBaseUrl).readText()
+        val downloadLink = URL(apiText.split("\"downloadLink\":\"")[1].split("\"")[0])
```
fsoffel, line 45: Is this approach to get the downloadLink safe? A proper JSON parsing would be. However, I think this is probably fine.

dataland-batch-manager/src/main/kotlin/org/dataland/datalandbatchmanager/service/GleifApiAccessor.kt
```diff
@@ -35,6 +37,17 @@ class GleifApiAccessor(
         downloadFileFromGleif("latest.csv", targetFile, "full Golden Copy File")
     }
 
+    /**
+     * Downloads the latest complete Lei-ISIN mapping file
+     * @param targetFile the local target file to be written
+     */
+    fun getIsinMappingFile(targetFile: File) {
```
fsoffel, line 43: What about `getFullIsinMappingFile` instead? This would be more clear about what is actually downloaded here and consistent with the download stuff regarding the gleif files

dataland-batch-manager/src/main/kotlin/org/dataland/datalandbatchmanager/service/GleifApiAccessor.kt
```diff
@@ -10,12 +10,14 @@ import java.io.FileNotFoundException
...
 import java.net.URL
 
+
 /**
  * The class to download the zipped GLEIF golden copy CSV files
  */
 @Component
 class GleifApiAccessor(
     @Value("\${gleif.download.baseurl}") private val gleifBaseUrl: String,
+    @Value("\${gleif.mapping.download.baseurl}") private val gleifMappingBaseUrl: String,
```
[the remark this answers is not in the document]
fsoffel: After realizing that this is only the url to the file that links the mapping file. This parameter and app prop should be named appropriately. Something like `gleifToIsinMappingReferenceUrl` is the best I came up with now.

dataland-batch-manager/src/main/kotlin/org/dataland/datalandbatchmanager/service/GleifGoldenCopyIngestor.kt
```diff
@@ -22,19 +22,21 @@ import kotlin.time.toDuration
...
     @Autowired private val gleifParser: GleifCsvParser,
     @Autowired private val companyUploader: CompanyUploader,
     @Autowired private val actuatorApi: ActuatorApi,
+    @Autowired private val isinDeltaBuilder: IsinDeltaBuilder,
     @Value("\${dataland.dataland-batch-managet.get-all-gleif-companies.force:false}")
     private val allCompaniesForceIngest: Boolean,
-
     @Value("\${dataland.dataland-batch-managet.get-all-gleif-companies.flag-file:#{null}}")
     private val allCompaniesIngestFlagFilePath: String?,
+    @Value("\${dataland.dataland-batch-manager.mapping-file}")
```
fsoffel, line 38: What happens here if there is no file behind this path? Is then only the exists flag set to false?

fsoffel: Did you test that this string to file conversion works?

dataland-batch-manager/src/main/kotlin/org/dataland/datalandbatchmanager/service/CompanyUploader.kt
```diff
@@ -126,4 +127,33 @@ class CompanyUploader(
             )
         }
     }
+
```
fsoffel, line 130: My comments in the following might become irrelevant. As I understood the ticket this upload procedure should have been incorporated into the already existing requests done in the functions above to reduce traffic. I don't know if this was only about the one off upload and not the sunday updates but let's clarify this. I feel like the ticket description was not super clear here again.

dataland-batch-manager/src/main/kotlin/org/dataland/datalandbatchmanager/service/CompanyUploader.kt
```diff
@@ -126,4 +127,33 @@ class CompanyUploader(
...
+    /**
+     * Updates the ISINs of all companies.
+     * @param leiIsinMap the delta-map with the format "LEI"->"ISIN1,ISIN2,..."
+     */
+    fun updateIsinMapping(leiIsinMap: Map<String, String>) {
```
fsoffel, line 135: I like `isinMapping`, however, we should really stay consistent with the naming! The parameter is called `leiIsinMap` I saw `gleifMapping` somewhere, I think, and only `mapping` as well e.g. in the application.properties.

dataland-batch-manager/src/main/kotlin/org/dataland/datalandbatchmanager/service/CompanyUploader.kt
```diff
@@ -126,4 +127,33 @@ class CompanyUploader(
...
+        retryOnCommonApiErrors {
+            for ((lei, isins) in leiIsinMap) {
+                val isinList = isins.split(",").map { it.trim() }
+                val company = companyDataControllerApi.getCompaniesBySearchString(lei)
```
fsoffel, line 139: I hate that we use this endpoint for this type of search. We should have a proper `GET /companies/identifiers/{identiferType}/identifier` endpoint. But nothing to do here.

fsoffel, line 139: *`companies`

dataland-batch-manager/src/main/kotlin/org/dataland/datalandbatchmanager/service/CompanyUploader.kt
```diff
@@ -126,4 +127,33 @@ class CompanyUploader(
...
+                if (company.isNotEmpty()) {
```
fsoffel, line 140: We should actually check that there is exactly one result. If there are more the first one is not necessarily the one we are looking for.

fsoffel: Hold on, I think it would be best to filter `company`/`companies` for something with the required lei and check if that list ist not empty and take the first from these result, because you do not check for the identifier to be the correct LEI, yet anyways.

dataland-batch-manager/src/main/kotlin/org/dataland/datalandbatchmanager/service/CompanyUploader.kt
```diff
@@ -126,4 +127,33 @@ class CompanyUploader(
...
+                    val updatedIdentifiers = existingIdentifiers.toMutableMap()
+                    updatedIdentifiers["isin"] = isinList
+
+                    val companyPatch = CompanyInformationPatch(identifiers = updatedIdentifiers)
+                    companyDataControllerApi.patchCompanyById(
+                        companyId,
+                        companyPatch,
+                    )
+                } else {
+                    logger.info("No company found for LEI: $lei")
```
fsoffel, line 154: Should this rather be an exception? Because in this case we made an erro rsomewhere when importing the companies from the original GLEIF ingestion.

dataland-batch-manager/src/main/kotlin/org/dataland/datalandbatchmanager/service/CompanyUploader.kt
fsoffel, line 1: Regarding the `retryOnCommonApiErrors` function:
We should definitely throw/forward an error if max retries was reached! Because, as I understood, this should not happen.

dataland-batch-manager/src/main/kotlin/org/dataland/datalandbatchmanager/service/GleifGoldenCopyIngestor.kt
```diff
@@ -65,22 +67,33 @@ class GleifGoldenCopyIngestor(
 
             logger.info("Retrieving all company data available via GLEIF.")
             val tempFile = File.createTempFile("gleif_golden_copy", ".csv")
-            processFile(tempFile, gleifApiAccessor::getFullGoldenCopy)
+            processDeltaFile(tempFile, gleifApiAccessor::getFullGoldenCopy)
+            if (savedMappingFile.exists()) {
+                if (!savedMappingFile.delete()) {
```
fsoffel, line 72: We should not try to delete it because it should not be there in the first place. Also, rather throw an error if it already exists.

dataland-batch-manager/src/main/kotlin/org/dataland/datalandbatchmanager/service/GleifGoldenCopyIngestor.kt
```diff
@@ -94,6 +107,31 @@ class GleifGoldenCopyIngestor(
...
+        processMappingFile(tempFile, gleifApiAccessor::getIsinMappingFile)
+    }
+
+    @Synchronized
+    private fun processMappingFile(newMappingFile: File, downloadFile: (file: File) -> Unit) {
+        waitForBackend()
+        val start = System.nanoTime()
+
+        downloadFile(newMappingFile)
+        val newMappingFileCsv = gleifParser.getCsvFileFromZip(newMappingFile)
```
fsoffel, line 122: Shouldn't this overwrite the savedMappingFile at some point? If that exists yet, where does that happen?

dataland-batch-manager/src/main/kotlin/org/dataland/datalandbatchmanager/service/IsinDeltaBuilder.kt
```diff
@@ -0,0 +1,116 @@
...
+        return deltaMapping
+    }
+
+    /**
+     * Compares the two LEI-ISIN maps and identifies LEIs that have new ISINs
+     * @param newMap the new LEI-ISIN map
+     * @param oldMap the old LEI-ISIN map
+     * @return map of changed LEI-ISINs
+     */
+    fun findLeisWithUpdatedIsin(newMap: Map<String, String>, oldMap: Map<String, String>): Map<String, String> {
```
fsoffel, line 44: Could be private

dataland-batch-manager/src/main/kotlin/org/dataland/datalandbatchmanager/service/IsinDeltaBuilder.kt
```diff
@@ -0,0 +1,116 @@
...
+    }
+
+    /**
+     * Replaces the locally saved old mapping file with the recently downloaded one after creating delta is done
+     * @param newMappingFile latest version of the LEI-ISIN mapping file
+     */
+    fun replaceOldMappingFile(newMappingFile: File) {
+        try {
+            if (savedMappingFile.exists() && !savedMappingFile.delete()) {
+                logger.error("Unable to delete the old mapping file $savedMappingFile")
```
fsoffel, line 106: Throw an error. I also think deletion is not necessary?

fsoffel: Also I find this function should not be in this class.

dataland-batch-manager/src/main/kotlin/org/dataland/datalandbatchmanager/service/IsinDeltaBuilder.kt
```diff
@@ -0,0 +1,116 @@
...
+                return
+            }
+            val renamedFile = File(savedMappingFile.parent, "isinMapping.csv")
```
fsoffel, line 109: Do we need to generate this string `renamedFile` can we not directly copy to savedMappingFile?

dataland-batch-manager/src/main/kotlin/org/dataland/datalandbatchmanager/service/IsinDeltaBuilder.kt
```diff
@@ -0,0 +1,116 @@
...
+            newMappingFile.copyTo(renamedFile, true)
+            newMappingFile.delete()
+        } catch (e: FileSystemException) {
+            logger.error("Error while replacing the old mapping file: ${e.message}")
```
fsoffel, line 113: Why catch the exception. Throw it.

dataland-batch-manager/src/main/kotlin/org/dataland/datalandbatchmanager/service/IsinDeltaBuilder.kt
```diff
@@ -0,0 +1,116 @@
...
+    private val logger = LoggerFactory.getLogger(javaClass)
+
+    /**
+     * Creates the delta file of the updated LEI-ISIN mapping file
+     * @param newMappingFile latest version of the LEI-ISIN mapping file
+     * @param oldMappingFile previously stored version of the LEI-ISIN mapping file, if one exists
+     */
+    fun createDeltaOfMappingFile(newMappingFile: File, oldMappingFile: File?): Map<String, String> {
+        val newMapping = parseCsvToGroupedMap(newMappingFile)
+        if (oldMappingFile == null) {
```
fsoffel, line 28: the function name does not suggest that file movement happens.

dataland-batch-manager/src/main/kotlin/org/dataland/datalandbatchmanager/service/IsinDeltaBuilder.kt
```diff
@@ -0,0 +1,116 @@
...
+     * Coverts CSV file to a LEI-ISIN map, while also aggregating all ISINs of a specific
+     * LEI into one comma-separated string
+     * @param csvFile the file to be parsed
+     * @return map of LEI-ISINs
+     */
+    private fun parseCsvToGroupedMap(csvFile: File): Map<String, String> {
+        val csvMapper = CsvMapper()
+        csvMapper.registerModule(kotlinModule())
+
+        val csvSchema = CsvSchema.builder()
```
fsoffel, line 67: This csv does not have multiple columns. It only has one. Therefore this would not work.

dataland-batch-manager/src/main/kotlin/org/dataland/datalandbatchmanager/service/IsinDeltaBuilder.kt
```diff
@@ -0,0 +1,116 @@
...
+            val csvParser = csvMapper
+                .readerFor(Map::class.java)
+                .with(csvSchema)
+                .readValues<Map<String, String>>(csvFile)
+
+            csvParser.forEach { entry ->
+                val lei = entry["LEI"]
+                val isin = entry["ISIN"]
+
+                if (lei != null && isin != null && mappings.containsKey(lei)) {
```
fsoffel, line 85: What is going on here? Might there be multiple entries for each lei? If we do it like this, would each mapping[x] not start with a ",". And if we split this into the list later on would the first lei not be an empty string?

dataland-batch-manager/src/main/kotlin/org/dataland/datalandbatchmanager/service/IsinDeltaBuilder.kt
```diff
@@ -0,0 +1,116 @@
...
+                    mappings[lei]?.append(",")
```
fsoffel, line 86: Why not directly store a proper list of isins as `List<String>` instead of `String`?

dataland-batch-manager/src/main/kotlin/org/dataland/datalandbatchmanager/service/IsinDeltaBuilder.kt
```diff
@@ -0,0 +1,116 @@
...
+                    mappings[lei]?.append(isin)
+                } else if (lei != null && isin != null && !mappings.containsKey(lei)) {
+                    mappings[lei] = StringBuilder(isin)
+                }
+            }
+        } catch (e: FileSystemException) {
+            logger.error("Error while parsing CSV: ${e.message}")
```
fsoffel, line 93: Why catch that error? Throw it. This should not happen in a production scenario.

dataland-batch-manager/src/main/kotlin/org/dataland/datalandbatchmanager/service/IsinDeltaBuilder.kt
```diff
@@ -0,0 +1,116 @@
...
+     * @param oldMap the old LEI-ISIN map
+     * @return map of changed LEI-ISINs
+     */
+    fun findLeisWithUpdatedIsin(newMap: Map<String, String>, oldMap: Map<String, String>): Map<String, String> {
+        val deltaMap = mutableMapOf<String, String>()
+
+        for ((lei, newIsins) in newMap) {
+            val oldIsins = oldMap[lei]
+
+            if (oldIsins == null || oldIsins != newIsins) {
```
fsoffel, line 50: What if oldIsins and newIsins only differ in order? If they would both be Set<String> this would not be a problem. There is nothing speaking against implementing them like that.

aliabdoli77: The csv file is ordered alphabetically so a mere reordering is unlikely. and even if that happens, nothing horrible happens.

dataland-batch-manager/src/main/kotlin/org/dataland/datalandbatchmanager/service/CompanyUploader.kt
fsoffel: As clarified in today's daily this sequential implementation is perfectly fine.

dataland-batch-manager/src/main/kotlin/org/dataland/datalandbatchmanager/service/IsinDeltaBuilder.kt
aliabdoli77: I thought comparing two strings would be slightly faster than comparing two lists

.githooks/pre-commit
aliabdoli77: it was renamed once accidentally

dataland-batch-manager/src/main/resources/application.properties
aliabdoli77: done

dataland-batch-manager/src/main/resources/application.properties
aliabdoli77: done

dataland-batch-manager/src/main/resources/application-development.properties
aliabdoli77: removed it entirely

dataland-batch-manager/src/main/kotlin/org/dataland/datalandbatchmanager/service/GleifApiAccessor.kt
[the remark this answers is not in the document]
aliabdoli77: done

dataland-batch-manager/src/main/kotlin/org/dataland/datalandbatchmanager/service/GleifGoldenCopyIngestor.kt
```diff
@@ -65,22 +67,33 @@ class GleifGoldenCopyIngestor(
...
             logger.info("Flag file not present & no force update variable set => Not performing any download")
         }
     }
 
     @Suppress("UnusedPrivateMember") // Detect does not recognise the scheduled execution of this function
     @Scheduled(cron = "0 0 3 * * SUN")
-    private fun processDeltaFile() {
-        logger.info("Starting update cycle for latest delta file.")
+    private fun processUpdates() {
+        prepareDeltaFile()
```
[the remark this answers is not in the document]
aliabdoli77: done

dataland-batch-manager/src/main/kotlin/org/dataland/datalandbatchmanager/service/GleifApiAccessor.kt
aliabdoli77: done

dataland-batch-manager/src/main/kotlin/org/dataland/datalandbatchmanager/service/CompanyUploader.kt
aliabdoli77: done

dataland-batch-manager/src/main/kotlin/org/dataland/datalandbatchmanager/service/IsinDeltaBuilder.kt
fsoffel: Fair. Das ist glaube ich aber vernachlässigbar. Hoffe ich.

dataland-batch-manager/src/main/kotlin/org/dataland/datalandbatchmanager/service/CompanyUploader.kt
aliabdoli77: done

dataland-batch-manager/src/main/kotlin/org/dataland/datalandbatchmanager/service/GleifGoldenCopyIngestor.kt
aliabdoli77: Why not? What if the file exists due to last month's update, then somebody decides to activate the fullGoldenCopy.

we have this code later which will run much faster if we delete the savedMappingFile for a goldenCopy

val deltaMap: Map<String, String> = if (!savedMappingFile.exists() || savedMappingFile.length() == 0L) {
            isinDeltaBuilder.createDeltaOfMappingFile(newMappingFileCsv, null)
        } else {
            isinDeltaBuilder.createDeltaOfMappingFile(newMappingFileCsv, savedMappingFile)
        }

dataland-batch-manager/src/main/kotlin/org/dataland/datalandbatchmanager/service/GleifGoldenCopyIngestor.kt
aliabdoli77: replaceOldMappingFile in IsinDeltaBuilder.kt

dataland-batch-manager/src/main/kotlin/org/dataland/datalandbatchmanager/service/IsinDeltaBuilder.kt
aliabdoli77: done

dataland-batch-manager/src/main/kotlin/org/dataland/datalandbatchmanager/service/IsinDeltaBuilder.kt
aliabdoli77: deletion of savedMappingFile or newMappingFile or both?

dataland-batch-manager/src/main/kotlin/org/dataland/datalandbatchmanager/service/GleifCsvParser.kt
```diff
@@ -32,6 +33,22 @@ class GleifCsvParser {
         return BufferedReader(inputStreamReader)
     }
 
+    /**
+     * Extracts CSV file from Zip file
+     * @param zipFile the zip file
+     * @return CSV file inside Zip file
+     */
+    fun getCsvFileFromZip(zipFile: File): File {
```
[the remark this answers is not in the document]
aliabdoli77: relocated

dataland-batch-manager/src/main/kotlin/org/dataland/datalandbatchmanager/service/IsinDeltaBuilder.kt
fsoffel: Good point.

dataland-batch-manager/src/main/kotlin/org/dataland/datalandbatchmanager/service/IsinDeltaBuilder.kt
```diff
@@ -0,0 +1,112 @@
...
+
+            csvParser.forEach { entry ->
+                val lei = entry["LEI"]
+                val isin = entry["ISIN"]
+
+                if (lei != null && isin != null && mappings.containsKey(lei)) {
+                    mappings[lei]?.append(",")
+                    mappings[lei]?.append(isin)
+                }
+                if (lei != null && isin != null && !mappings.containsKey(lei)) {
```
fsoffel, line 89: Why remove is this no more else if?

dataland-batch-manager/src/main/kotlin/org/dataland/datalandbatchmanager/service/GleifApiAccessor.kt
```diff
@@ -46,7 +47,26 @@ class GleifApiAccessor(
...
+        FileUtils.copyFile(getCsvFileFromZip(tempZipFile), targetFile)
+        tempZipFile.delete()
+    }
+
+    /**
+     * Extracts CSV file from Zip file
+     * @param zipFile the zip file
+     * @return CSV file inside Zip file
+     */
+    fun getCsvFileFromZip(zipFile: File): File {
```
fsoffel, line 62: This function is not api related, still. But better here that in the csv parser, I guess. Can it be private though? I do not see the problem with having this in it's own class/object. Then we would not need to make it public for testing and responsibilities would be split more logically.

dataland-batch-manager/src/main/kotlin/org/dataland/datalandbatchmanager/service/GleifApiAccessor.kt
```diff
@@ -44,8 +44,9 @@ class GleifApiAccessor(
     fun getFullIsinMappingFile(targetFile: File) {
         val apiText = URL(gleifToIsinMappingReferenceUrl).readText()
         logger.info("API Response: $apiText")
-        //val downloadLink = URL(apiText.split("\"downloadLink\":\"")[1].split("\"")[0])
-        val downloadLink = URL("https://mapping.gleif.org/api/v2/isin-lei/465e4f8f-ade3-42f3-a014-e742a78e43df/download")
+        // val downloadLink = URL(apiText.split("\"downloadLink\":\"")[1].split("\"")[0])
+        val downloadLink =
+            URL("https://mapping.gleif.org/api/v2/isin-lei/465e4f8f-ade3-42f3-a014-e742a78e43df/download")
```
fsoffel, line 49: WIP

dataland-batch-manager/src/main/kotlin/org/dataland/datalandbatchmanager/service/CompanyUploader.kt
fsoffel: I think LEI is unique. However, the endpoint does not only search for LEI. I'm not sure if there could be a different identifier type with this LEI as value. This would be found by this endpoint as well. I just want to be safe that such an unlikely event can not happen.

dataland-batch-manager/src/main/kotlin/org/dataland/datalandbatchmanager/service/GleifGoldenCopyIngestor.kt
fsoffel: Isn't the file deleted anyways if we redeploy? And we can only trigger a new full copy at redeployment anyways. But I'm not sure about this.

fsoffel: Aah, you are preserving them via volumes. I need to further check that.

dataland-batch-manager/src/main/kotlin/org/dataland/datalandbatchmanager/service/IsinDeltaBuilder.kt
fsoffel: savedMappingFile in the if condition because it should be overwritten anyways on every happy path.

dataland-batch-manager/src/main/kotlin/org/dataland/datalandbatchmanager/service/IsinDeltaBuilder.kt
fsoffel: Maybe I understood that wrong. I thought we could do something like the following:
`newMappingFile.copyTo(savedMappingFile, true)`. Is this not where we want the new stuff?

dataland-batch-manager/src/main/kotlin/org/dataland/datalandbatchmanager/service/IsinDeltaBuilder.kt
aliabdoli77: done

dataland-batch-manager/src/main/kotlin/org/dataland/datalandbatchmanager/service/IsinDeltaBuilder.kt
aliabdoli77: done as discussed

dataland-batch-manager/src/main/kotlin/org/dataland/datalandbatchmanager/service/IsinDeltaBuilder.kt
aliabdoli77: removed. done

dataland-batch-manager/src/main/kotlin/org/dataland/datalandbatchmanager/service/IsinDeltaBuilder.kt
fsoffel: TODO for me: validate if the copy process works

dataland-batch-manager/src/main/kotlin/org/dataland/datalandbatchmanager/service/IsinDeltaBuilder.kt
aliabdoli77: done

dataland-batch-manager/src/test/kotlin/org/dataland/batchmanager/service/IsinDeltaBuilderTest.kt
```diff
@@ -0,0 +1,95 @@
...
+        newFile.deleteOnExit()
+    }
+
+    @Test
+    fun `test if delta of two files with LEI ISIN mapping looks as expected`() {
+        val isinDeltaBuilder = IsinDeltaBuilder()
+        assert(isinDeltaBuilder.createDeltaOfMappingFile(newFile, oldFile) == (deltaMap))
+    }
+
+//    @Test
```
fsoffel, line 76: ?

dataland-backend/src/main/kotlin/org/dataland/datalandbackend/api/CompanyApi.kt
```diff
@@ -146,6 +147,31 @@ interface CompanyApi {
         @PathVariable("identifier") identifier: String,
     )
 
+    /**
```
stewal19, line 150: This is the endpoint you have told me about in our call

dataland-batch-manager/src/main/kotlin/org/dataland/datalandbatchmanager/service/CompanyUploader.kt
```diff
@@ -62,7 +63,7 @@ class CompanyUploader(
         while (counter < MAX_RETRIES) {
             try {
                 functionToExecute()
-                break
+                return
```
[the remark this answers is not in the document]
fsoffel: Because otherwise the logging message in the end would be displazed in any case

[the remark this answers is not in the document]
stewal19: ok, makes sense

PR #397, opened 2024-01-10
Dala 2713 gdv voeb followup

fsoffel:
# Pull Request \<Title>
`<Description here>`
  - [ ] The version of main currently active on prod is deployed to a dev server with `Reset non-user related Docker Volumes & Re-populate` turned on
  - [ ] The data from prod is migrated using `./migrateData.sh dataland.com <SOURCE_API_KEY> <TARGET> <TARGET_API_KEY>` 
  - [ ] The feature branch is deployed to the same server with `Reset non-user related Docker Volumes & Re-populate` turned off
- [ ] Make sure that pull requests in the development tools repository, which are connected to this pull requests, 
are merged


dataland-frontend/tests/e2e/utils/FakeFixtureUtils.ts
```diff
@@ -151,6 +151,14 @@ export class Generator {
     return faker.company.buzzNoun();
   }
 
+  randomParagraph(): string | null {
+    return this.valueOrNull(this.guaranteedParagraph());
+  }
+
+  guaranteedParagraph(): string {
+    return faker.lorem.paragraphs({ min: 1, max: 5 });
```
fsoffel, line 159: `faker.lorem.paragraphs` actually generates multiple paragagraphs. `faker.lorem.paragraph` however generates a paragraph with the provided number of sentences. Theres is also `faker.lorem.lines` which generates multiple lines. I think one of the latter is rather what you want here.

dataland-frontend/tests/e2e/utils/FakeFixtureUtils.ts
```diff
@@ -151,6 +151,14 @@ export class Generator {
     return faker.company.buzzNoun();
   }
 
+  randomParagraph(): string | null {
+    return this.valueOrNull(this.guaranteedParagraph());
+  }
+
+  guaranteedParagraph(): string {
```
fsoffel, line 158: This is the first time I see this guaranteed notation. We would usually call this `generateParagraph`.

fsoffel: Also, docs missing. But sonar will tell you as well.

dataland-frontend/tests/e2e/specs/search-company-framework-data/SearchCompaniesForFrameworkDataDropdownFilter.ts
```diff
@@ -51,12 +51,12 @@ describe("As a user, I expect the search functionality on the /companies page to
         "eq",
         getBaseUrl() +
           "/companies?" +
-          `framework=${DataTypeEnum.Gdv}` +
```
fsoffel, line 54: Do you know what controls the order of these? Is it the enum? Why is the DataTypeEnum ordered like this?

dataland-frontend/tests/e2e/utils/FakeFixtureUtils.ts
```diff
@@ -151,6 +151,14 @@ export class Generator {
     return faker.company.buzzNoun();
   }
 
+  randomParagraph(): string | null {
```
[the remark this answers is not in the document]
emanueldfine: Not necessary here.  The other function also don't have docs in this file.

dataland-frontend/tests/e2e/utils/FakeFixtureUtils.ts
emanueldfine: Docs are not enforced here and other functions also don't have docs here.

emanueldfine: The other functions (which also exist in main) all use "guaranteed" here.   Do you mean to change them all?

fsoffel: Don't mind the first comment.

dataland-framework-toolbox/inputs/gdv/dataDictionary-GDV-VOEB-GDV-VÖB ESG questionnaire.csv
```diff
@@ -1,13 +1,13 @@
...
+2.2,Allgemein,ESG Ziele,Investitionen in Zielerreichung,Bitte geben Sie an wieviele Budgets/Vollzeitäquivalente für das Erreichen der ESG-Ziele zugewiesen wurden.,FreeText,,,,2,Yes,
 3,Allgemein,Sektoren,Sektoren mit hohen Klimaauswirkungen,Kann das Unternehmen einem oder mehreren Sektoren mit hohen Klimaauswirkungen zugeordnet werden?,Yes/No,,,,0,Yes,
-3.1,Allgemein,Sektoren,Auflistung der Sektoren,"Bitte geben Sie an, zu welchen Sektoren (mit hohen Klimaauswirkungen) das Unternehmen zugeordnet werden kann.",Multi-Select Dropdown,A; B; C; D; E; F; G; H; L,,,3,Yes,
+3.1,Allgemein,Sektoren,Auflistung der Sektoren,"Bitte geben Sie an, zu welchen Sektoren (mit hohen Klimaauswirkungen) das Unternehmen zugeordnet werden kann.",Multi-Select Dropdown,"A - Landwirtschaft, Forstwirtschaft und Fischerei| B - Bergbau und Gewinnung von Steinen und Erden|C - Verarbeitendes Gewerbe / Herstellung von Waren|D - Energieversorgung|E - Wasserversorgung; Abwasser & Abfallentsorgung; Beseitigungen von Umweltverschmutzungen|F - Baugewerbe/Bau|G - Handel; Instandhaltung und Reparatur von Kraftfahrzeugen|H - Verkehr und Lagerhaltung| L - Grundstücks- und Wohnungswesen",,,3,Yes,
```
fsoffel, line 8: There are spaces before some of the elements.

emanueldfine: That should not do a difference, because the function `trim()` is being used in the devtools on each option, and it removes spaces at the beginning and end of the string

dataland-framework-toolbox/src/main/kotlin/org/dataland/frameworktoolbox/specific/uploadconfig/elements/UploadCellConfigBuilder.kt
```diff
@@ -11,7 +11,7 @@ import org.dataland.frameworktoolbox.specific.viewconfig.functional.FrameworkBoo
  */
 
 @Suppress("LongParameterList")
-class CellConfigBuilder(
+class UploadCellConfigBuilder(
```
fsoffel, line 14: Why did you feel like renaming these classes and functions by giving them the "Upload" prefix?

dataland-framework-toolbox/src/main/kotlin/org/dataland/frameworktoolbox/frameworks/gdv/GdvFramework.kt
```diff
@@ -20,21 +20,42 @@ import org.dataland.frameworktoolbox.specific.uploadconfig.functional.FrameworkU
...
-class GdvFramework : InDevelopmentPavedRoadFramework(
+class GdvFramework : PavedRoadFramework(
     identifier = "gdv",
     label = "GDV/VÖB",
     explanation = "Das GDV/VÖB Framework ist ein ESG-Fragebogen des Gesamtverbands der Versicherer und des " +
         "Bundesverbands Öffentlicher Banken",
     File("./dataland-framework-toolbox/inputs/gdv/dataDictionary-GDV-VOEB-GDV-VÖB ESG questionnaire.csv"),
 ) {
 
+    override fun configureDiagnostics(diagnosticManager: DiagnosticManager) {
```
fsoffel, line 39: Are these errors still logged but not leading to failure? For me: To be tested

dataland-framework-toolbox/src/main/kotlin/org/dataland/frameworktoolbox/frameworks/InDevelopmentPavedRoadFramework.kt
```diff
@@ -31,44 +23,6 @@ abstract class InDevelopmentPavedRoadFramework(
 ) :
     PavedRoadFramework(identifier, label, explanation, frameworkTemplateCsvFile) {
 
-    override fun convertExcelTemplateToToHighLevelComponentRepresentation(
```
fsoffel, line 34: The corresponding method in the `PavedRoadFramework` class does no longer need to be open. Or should it be for maximum flexibility?

fsoffel: Looking on what this function is doing. I feel like we should not delete it. What if we are implementing another framework inheriting from `InDevelopmentPavedRoadFramework` we would not notice the issue, right? Is there a different mechanism in place which notifies us about this.

dataland-framework-toolbox/src/main/kotlin/org/dataland/frameworktoolbox/specific/uploadconfig/elements/UploadCellConfigBuilder.kt
emanueldfine: Because the "view-page"-generation also has a similar class which is also named `CellConfigBuilder`.
This makes it difficult to click through the code (sometimes you are confused at which one you are actually looking at).
In order to make it easier for code-readers to understand which CellConfigBuilder they are looking at, I decided to rename this one slightly different.

dataland-framework-toolbox/src/main/kotlin/org/dataland/frameworktoolbox/frameworks/gdv/GdvFramework.kt
emanueldfine: I'd expect them not to be logged anymore.

The background of this is, that these are fields with some "Custom ***" entry in the "Component" column.
We do not want the devtools to parse these fields. We want to manually create them in the customization hook.
In order to make the devtools not say "hey, wtf is Custom*** ?  I don't know such a component" we need these ignores here.

dataland-framework-toolbox/src/main/kotlin/org/dataland/frameworktoolbox/frameworks/InDevelopmentPavedRoadFramework.kt
emanueldfine: I am not sure. Let's have a call about this.

dataland-frontend/tests/e2e/utils/FakeFixtureUtils.ts
emanueldfine: Ok I looked at all three functions (`.paragraphs`, `.paragraph` and also `.lines`) and what they do.
Now I think `.paragraphs` is indeed the best thing to use here.

Reason:

We want to generate very long texts that also have line breaks.
`.lines()`does line breaks, but it only writes one short sentence per line. We want more text than that.
`.paragraph()` has multiple sentences, but it does not do line breaks.

However, `.paragraphs()` does line breaks AND it also writes a lot of sentences per line.
So it is exactly what we want here.


Still I would suggest to rename the function to `guranteedParagraphs` with a "s" at the end.

Is that ok with you?

dataland-framework-toolbox/src/main/kotlin/org/dataland/frameworktoolbox/intermediate/group/ComponentGroupApi.kt
```diff
@@ -70,10 +76,18 @@ inline fun <reified T : ComponentBase> ComponentGroupApi.delete(identifier: Stri
     return this.delete(identifier, T::class)
 }
 
-/**
+/**getOrNull
```
fsoffel, line 79: What is this here?

dataland-framework-toolbox/src/main/kotlin/org/dataland/frameworktoolbox/template/components/MultiSelectComponentFactory.kt
```diff
@@ -21,7 +21,6 @@ class MultiSelectComponentFactory(@Autowired val templateDiagnostic: TemplateDia
         utils: ComponentGenerationUtils,
         componentGroup: ComponentGroupApi,
     ): ComponentBase {
-        templateDiagnostic.optionsNotUsed(row)
```
fsoffel, line 24: Why was this even there?

dataland-framework-toolbox/src/main/kotlin/org/dataland/frameworktoolbox/template/components/SingleSelectComponentFactory.kt
```diff
@@ -21,7 +21,6 @@ class SingleSelectComponentFactory(@Autowired val templateDiagnostic: TemplateDi
         utils: ComponentGenerationUtils,
         componentGroup: ComponentGroupApi,
     ): ComponentBase {
-        templateDiagnostic.optionsNotUsed(row)
```
fsoffel, line 24: Why was this even there?

dataland-framework-toolbox/src/main/kotlin/org/dataland/frameworktoolbox/intermediate/components/IntegerComponent.kt
```diff
@@ -0,0 +1,68 @@
...
+    identifier: String,
+    parent: FieldNodeParent,
+) : ComponentBase(identifier, parent) {
+    open var constantUnitSuffix: String? = null
+
+    override fun generateDefaultDataModel(dataClassBuilder: DataClassBuilder) {
+        dataClassBuilder.addProperty(
+            this.identifier,
+            documentSupport.getJvmTypeReference(
+                TypeReference("java.math.BigInteger", isNullable),
```
fsoffel, line 26: For `DecimalComponent` the value `isNullable` is not passed into `TypeReference(...)` but the constant `true`. Why is this different here?

dataland-framework-toolbox/src/main/kotlin/org/dataland/frameworktoolbox/intermediate/components/IntegerComponent.kt
```diff
@@ -0,0 +1,68 @@
...
+            ),
+        )
+    }
+
+    override fun generateDefaultUploadConfig(uploadCategoryBuilder: UploadCategoryBuilder) {
+        uploadCategoryBuilder.addStandardUploadConfigCell(
+            component = this,
+            uploadComponentName = "NumberFormField",
+            validation = "integer",
+            unit = constantUnitSuffix,
```
fsoffel, line 54: Why is this unit field not also set in the DecimalComponent?

dataland-framework-toolbox/src/main/kotlin/org/dataland/frameworktoolbox/intermediate/components/IntegerComponent.kt
```diff
@@ -0,0 +1,68 @@
...
+import org.dataland.frameworktoolbox.specific.viewconfig.functional.FrameworkDisplayValueLambda
+
+/**
+ * A IntegerComponent represents a numeric value from the integer domain
+ */
+open class IntegerComponent(
+    identifier: String,
+    parent: FieldNodeParent,
+) : ComponentBase(identifier, parent) {
+    open var constantUnitSuffix: String? = null
```
fsoffel, line 20: `DecimalComponent` has `(minimum|maximum)Value` fields. Why not this one? If not provided via template. They definitely should be.

dataland-framework-toolbox/src/main/kotlin/org/dataland/frameworktoolbox/intermediate/components/IntegerComponent.kt
fsoffel, line 1: I feel like this class and `DecimalComponent` might have a different parent class which reduces the similarities between these two classes.

dataland-framework-toolbox/src/main/kotlin/org/dataland/frameworktoolbox/template/components/IntegerComponentFactory.kt
```diff
@@ -0,0 +1,43 @@
...
+@Component
+class IntegerComponentFactory(@Autowired val templateDiagnostic: TemplateDiagnostic) : TemplateComponentFactory {
+    override fun canGenerateComponent(row: TemplateRow): Boolean = row.component == "Integer"
+
+    override fun generateComponent(
+        row: TemplateRow,
+        utils: ComponentGenerationUtils,
+        componentGroup: ComponentGroupApi,
+    ): ComponentBase {
+        templateDiagnostic.optionsNotUsed(row)
```
fsoffel, line 24: Again, why do we not use units for decimals in the corresponding factory? Just because there is no use case?

dataland-framework-toolbox/src/main/kotlin/org/dataland/frameworktoolbox/template/components/ComponentFactoryContainer.kt
```diff
@@ -0,0 +1,10 @@
+package org.dataland.frameworktoolbox.template.components
+
+import org.springframework.beans.factory.annotation.Autowired
+import org.springframework.stereotype.Component
+
+/**
+ * A surrogate component for retrieving an autowired ordered list of TemplateComponentFactories
+ */
+@Component
+class ComponentFactoryContainer(@Autowired val factories: List<TemplateComponentFactory>)
```
fsoffel, line 10: Why do we use this wrapper? If we can autowire the list of `TemplateComponentFactory`s, why does the bean exist in the first place?

dataland-framework-toolbox/src/main/kotlin/org/dataland/frameworktoolbox/frameworks/PavedRoadFramework.kt
```diff
@@ -69,7 +69,8 @@ abstract class PavedRoadFramework(
     open fun getComponentFactoriesForIntermediateRepresentation(
         context: ApplicationContext,
     ): List<TemplateComponentFactory> {
-        return context.getBeansOfType<TemplateComponentFactory>().values.toList()
+        val containerBean = context.getBean<ComponentFactoryContainer>()
+        return containerBean.factories.reversed()
```
fsoffel, line 73: Why reversed? Where is the order controlled before this?

dataland-framework-toolbox/src/main/kotlin/org/dataland/frameworktoolbox/frameworks/PavedRoadFramework.kt
```diff
@@ -166,12 +167,7 @@ abstract class PavedRoadFramework(
         val dataModel = generateDataModel(framework)
         customizeDataModel(dataModel)
 
-        @Suppress("TooGenericExceptionCaught")
-        try {
-            dataModel.build(into = datalandProject)
-        } catch (ex: Exception) {
-            logger.error("Could not build framework data-model!", ex)
-        }
+        dataModel.build(into = datalandProject)
```
fsoffel, line 170: Why this change?

dataland-framework-toolbox/src/main/kotlin/org/dataland/frameworktoolbox/frameworks/PavedRoadFramework.kt
fsoffel: To get the warning factory to the end I guess? The rest of the order is somewhat arbitrary?

dataland-framework-toolbox/src/main/kotlin/org/dataland/frameworktoolbox/intermediate/group/ComponentGroupApi.kt
emanueldfine: Removed it.

dataland-framework-toolbox/src/main/kotlin/org/dataland/frameworktoolbox/template/components/MultiSelectComponentFactory.kt
emanueldfine: That must have been a copy-paste-leftover by me. 
Of course it does not make sense to assert that the options column is empty for a Multi-Select field row in the excel.
You need to paste values there to pass the options to the field.

dataland-framework-toolbox/src/main/kotlin/org/dataland/frameworktoolbox/template/components/SingleSelectComponentFactory.kt
emanueldfine: Same answer as for the Multi-Select component. Copy-paste leftover

dataland-framework-toolbox/src/main/kotlin/org/dataland/frameworktoolbox/intermediate/components/IntegerComponent.kt
emanueldfine: I think noone noticed that it is missing in the DecimalComponent because there is no decimal field in this framework which has a unit.
But it makes sense to add the possibility to add a unit there too, right?

Shall I add it there too? What do you think? Personally I think we should do it.

dataland-framework-toolbox/src/main/kotlin/org/dataland/frameworktoolbox/intermediate/components/IntegerComponent.kt
emanueldfine: I have no idea.  I think we should also write `isNullable` in the DecimalComponent.
Otherwise you cannot make decimal fields non-nullable via the excel column "Manadtory field".

What do you think?

dataland-framework-toolbox/src/main/kotlin/org/dataland/frameworktoolbox/intermediate/components/IntegerComponent.kt
emanueldfine: Ok but I just checked the code base, and it turns out that `minimumValue` is only used in a test, and `maximumValue` is not used at all.
So we could also remove these class variables without losing anything.

The other option would be to include it in the `IntegerComponent` too, but then we should also actually do smth with it, like e.g. passing a validator for this into the UploadConfig, or generating the fixtures accordingly.
Furthermore we would need the possibility to pass min and max values via the excel template.
However this sounds like it's not done in 10min.   So should we rather do it in some other ticket?

dataland-framework-toolbox/src/main/kotlin/org/dataland/frameworktoolbox/intermediate/components/IntegerComponent.kt
emanueldfine: You mean something like "NumberBaseComponent" which holds the overlap between "DecimalComponent" and "IntegerComponent"?
I could do that.

dataland-framework-toolbox/src/main/kotlin/org/dataland/frameworktoolbox/template/components/IntegerComponentFactory.kt
emanueldfine: We probably did not notice because there simply is no decimal field with a unit attached to it in the excel of Gdv.

fsoffel: Then we will added when the time has come.

dataland-framework-toolbox/src/main/kotlin/org/dataland/frameworktoolbox/intermediate/components/IntegerComponent.kt
fsoffel: That would make sense in my view because of the shared properties.

dataland-framework-toolbox/src/main/kotlin/org/dataland/frameworktoolbox/intermediate/components/IntegerComponent.kt
fsoffel: Yup. Let's not do this here. But maybe remove it in `DecimalComponent` then. I'll leave this up to you. I'll just let this conversation open till the end of the day for you to have the chance to read this answer.

dataland-framework-toolbox/src/main/kotlin/org/dataland/frameworktoolbox/intermediate/components/IntegerComponent.kt
fsoffel: I think we are less prone ot bugs if unit is also set there too. Maybe a shared base class can handle this.

dataland-framework-toolbox/src/main/kotlin/org/dataland/frameworktoolbox/intermediate/components/IntegerComponent.kt
fsoffel: I would like to talk about that. In this version you can not control the datapoint to be nullable and the template type to be nullable independently. Maybe I just don't understand something correclty.

dataland-framework-toolbox/src/main/kotlin/org/dataland/frameworktoolbox/template/components/MultiSelectComponentFactory.kt
fsoffel: And we just overllooked this because we used the indevpavedroadframework? With the pavedroadframework we would have noticed that right?

dataland-framework-toolbox/src/main/kotlin/org/dataland/frameworktoolbox/specific/uploadconfig/elements/UploadCellConfigBuilder.kt
fsoffel: I see. In the long run we should rename the view page versions similarly.

fsoffel: To be closed.

dataland-frontend/tests/e2e/utils/FakeFixtureUtils.ts
fsoffel: Agreed

dataland-framework-toolbox/src/main/kotlin/org/dataland/frameworktoolbox/frameworks/gdv/GdvFramework.kt
fsoffel: TODO for me. Check this function further.

dataland-framework-toolbox/src/main/kotlin/org/dataland/frameworktoolbox/frameworks/PavedRoadFramework.kt
fsoffel: But why was the warning added for this framework as well.

dataland-frontend/tests/e2e/fixtures/frameworks/gdv/GdvPreparedFixtures.ts
```diff
@@ -87,12 +87,23 @@ function modifyPreparedFixturesPartTwo(newFixture: FixtureData<GdvData>): void {
   if (newFixture.t.allgemein?.anleihen?.sustainibilityLinkedDebt) {
     newFixture.t.allgemein.anleihen.sustainibilityLinkedDebt = YesNo.Yes;
   }
+  if (newFixture.t.allgemein?.esgBerichte?.aktuelleBerichte) {
```
fsoffel, line 90: Why are these only conditionally added.

dataland-frontend/tests/e2e/fixtures/frameworks/gdv/GdvPreparedFixtures.ts
```diff
@@ -122,3 +133,22 @@ function modifyPreparedFixturesPartTwo(newFixture: FixtureData<GdvData>): void {
...
+ * Creates the contents for a time series based on a current year value and sets null for all numbers
+ * @param currentYear is the base year for the time series
+ * @returns the map of years to time series data objects
+ */
+function createEnergieverbrauchWithNullNumbers(currentYear: number): {
+  [p: string]: BerichterstattungEnergieverbrauchValues;
+} {
+  const objectToReturn: { [p: string]: BerichterstattungEnergieverbrauchValues } = {};
+  for (let i = -3; i <= 3; i++) {
+    objectToReturn[(currentYear + i).toString()] = {
```
fsoffel, line 147: In the backend it gets mapped to a map<Int, whatever> but here the keys here are strings. Is this a problem when posting the data?

dataland-framework-toolbox/inputs/gdv/dataDictionary-GDV-VOEB.xlsx
fsoffel, line 1: To be removed?

dataland-framework-toolbox/src/main/kotlin/org/dataland/frameworktoolbox/template/components/MultiSelectComponentFactory.kt
emanueldfine: Yes I think so.  I think the PavedRoadFramework would have stopped the run at this point.
The InDevelopment... just logs warnings and continues.

dataland-framework-toolbox/src/main/kotlin/org/dataland/frameworktoolbox/intermediate/components/IntegerComponent.kt
emanueldfine: I think removing it makes sense. It is dead code.

So pls leave this open until I have removed it.
I will notify you then.

emanueldfine: Removed it and also adjusted the test which used "minimumValue". See the commit

dataland-frontend/src/utils/Constants.ts
```diff
@@ -36,3 +37,23 @@ export const REGEX_FOR_FILE_NAMES = /^[^<>:"|?/*\\\s][^<>:"|?/*\\]{0,252}[^<>:"|
 export const MS_PER_DAY = 24 * 60 * 60 * 1000;
 
 export const NO_DATA_PROVIDED = "No data provided";
+
+/**
+ * Changes the sorting of a list of data type enums by putting the gdv/vöb framework at the very end.
+ * @param frameworksToInclude a unsorted list of data type enums
+ * @returns the list of data type enums sorted in a way, that gdv/vöb framework is the last element
+ */
+export function putGdvAtTheEndOfList(frameworksToInclude: DataTypeEnum[]): DataTypeEnum[] {
```
fsoffel, line 46: TODO for me: test if this works as expected

dataland-frontend/src/utils/DataSource.ts
```diff
@@ -0,0 +1,19 @@
...
+ */
+export function isValidFileName(
+  isMounted: boolean,
+  fileName: string | undefined,
+  invalidFileNames = [noReportLabel],
+): boolean {
+  if (!isMounted) {
+    return true;
+  }
+  return !!fileName && fileName?.length > 0 && !invalidFileNames.includes(fileName);
```
fsoffel, line 16: Doesn't `!!filename` render the second check redundant because the second check is only false if `filename == ""` but then `!!filename` is false.

dataland-frontend/src/utils/DataSource.ts
```diff
@@ -0,0 +1,19 @@
+/**
+ * Checks whether the Assurance data source has appropriate values
```
fsoffel, line 2: Docstring is referring to Assurance.

dataland-frontend/src/components/resources/dataTable/modals/GdvListOfBaseDataPointModal.vue
```diff
@@ -29,10 +29,10 @@ const tableData = computed(() => {
     <Column :header="dialogData.documentColumnHeader" headerStyle="width: 15vw;">
       <template #body="{ data }">
         <DocumentLink
+          :show-icon="!(data.dataSource == undefined)"
```
fsoffel, line 32: Shouldn't this rather be a show if on the DocumentLink? I mean effectively it makes no difference.

dataland-frontend/src/components/forms/parts/fields/StringBaseDataPointFormField.vue
```diff
@@ -1,16 +1,20 @@
 <template>
   <div class="form-field">
-    <FormKit v-model="baseDataPoint" type="group" name="dataSource">
+    <FormKit v-model="baseDataPoint" type="group" ignore="true">
```
fsoffel, line 3: Does the ignore flag work this way? Mustn't it be `:ignore="true"` or just `ignore`?

fsoffel: Must this even be a formkit now? There is no FormkitField inside.

fsoffel: This v-model seems unused to me as well.

dataland-frontend/src/components/forms/parts/fields/StringBaseDataPointFormField.vue
```diff
@@ -1,16 +1,20 @@
...
         name="name"
         :more-than-one-document-allowed="false"
         :file-names-for-prefill="fileNamesForPrefill"
       />
+    </FormKit>
+
+    <FormKit v-if="isValidFileName(isMounted, documentName)" type="group" name="dataSource">
       <FormKit type="hidden" name="fileName" v-model="documentName" />
-      <FormKit type="hidden" name="fileReference" v-model="documentReference" />
+      <FormKit type="hidden" name="fileReference" :modelValue="documentReference" />
```
fsoffel, line 15: Why use `modelValue` for `fileReference` but not for `fileName`? They basically work the same.

dataland-frontend/src/components/forms/parts/fields/StringBaseDataPointFormField.vue
```diff
@@ -41,12 +46,15 @@ export default defineComponent({
...
       fileNamesForPrefill: [] as string[],
       isMounted: false,
+      isValidFileName: isValidFileName,
     };
   },
   emits: ["fieldSpecificDocumentsUpdated"],
   mounted() {
-    this.updateFileUploadFiles();
-    this.isMounted = true;
+    setTimeout(() => {
```
fsoffel, line 54: This is to execute this only in the next event cycle? Can we talk about this?

dataland-framework-toolbox/src/main/kotlin/org/dataland/frameworktoolbox/intermediate/components/IntegerComponent.kt
emanueldfine: I will create a NumberBaseComponent.
PercentageComponent, IntegerComponent and also DecimalComponent will inherit from it.
By doing this, I could remove a lot of duplicate code (overlaps) among all these three components.

I will notify you when this is ready

dataland-frontend/src/components/resources/dataTable/modals/GdvListOfBaseDataPointModal.vue
emanueldfine: Yes I thought about it, but then I saw that DocumentLink already has a "show-icon" prop to control the display of the icon.
But I can also put the whole thing into a v-if if you want.   For me both is ok.
What do you think?

dataland-frontend/src/components/pages/CompanyCockpitPage.vue
```diff
@@ -5,7 +5,7 @@
     <div class="card-wrapper">
       <div class="card-grid">
         <FrameworkSummaryPanel
-          v-for="framework of Object.values(DataTypeEnum)"
+          v-for="framework of ARRAY_OF_FRAMEWORKS_WITH_VIEW_PAGE"
```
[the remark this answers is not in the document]
emanueldfine: No strong opinion here.
I could also use `ARRAY_OF_SUPPORTED_FRAMEWORKS` here instead.
Then also API-only frameworks will at least be displayed here (even though you cannot do anything with it).

What do you think?

dataland-frontend/src/components/forms/parts/fields/StringBaseDataPointFormField.vue
emanueldfine: Not my code.
Git says @pioSko2

dataland-frontend/src/components/forms/parts/fields/StringBaseDataPointFormField.vue
emanueldfine: Not my code.
Git says @pioSko2

dataland-frontend/src/components/forms/parts/fields/StringBaseDataPointFormField.vue
emanueldfine: Not my code.
Git says @pioSko2

dataland-frontend/src/components/resources/dataTable/MultiLayerDataTableBody.vue
```diff
@@ -110,6 +125,10 @@ function expandSectionsOnPageLoad(): void {
   }
 }
 
+const columnWidthStyle = computed(() => {
```
fsoffel, line 128: Won't this look very condensed if there are datasets for many reporting periods? Don't we have something like sideways scrolling?

dataland-frontend/src/utils/DataSource.ts
emanueldfine: Not my code.
Git says @pioSko2

dataland-frontend/src/utils/DataSource.ts
emanueldfine: Not my code.
Git says @pioSko2

dataland-framework-toolbox/inputs/gdv/dataDictionary-GDV-VOEB.xlsx
emanueldfine: I can remove it at the very end.
Let's keep it for dev purposes until then and keep this open as reminder, ok?

dataland-frontend/src/components/forms/parts/fields/StringBaseDataPointFormField.vue
pioSko2: Yes, mine. setTimeout is needed to create enough time for the page to load and populate dataSource inputs before the dataSource form is hidden. Without it, when editing, the dataSource is not filled in and re-sent empty

pioSko2: Will gladly discuss if you have another solution

dataland-frontend/src/utils/DataSource.ts
pioSko2: Yes, good point

dataland-frontend/src/components/forms/parts/fields/StringBaseDataPointFormField.vue
pioSko2: it is being used. Both this input and the hidden dataSource inputs use v-model in order to update each other. When the user uses the visible inputs, they update the hidden ones. When the user opens the Edit page, the hidden inputs are first populated (from the incoming data) and they in turn update the visible inputs visible to the user.

dataland-frontend/tests/e2e/fixtures/frameworks/gdv/GdvPreparedFixtures.ts
emanueldfine: No.
JS handles that map as an object like this:

``{ 
2021: ...
2022: ...
}``

So if I want to do an entry for 2023 into this "map", I need to do

`mapName["2023"] = ...`

For TS/JS the key is a string, and that is why I need to cast the number to a string here.


If you post this dataset and this object/JSON reaches our backend, it will deserialize this object to a `Map<Int, WhateverObject>`
As long as the key in the object is castable to an Int for Kotlin, you will have no problems.

dataland-frontend/src/components/resources/dataTable/MultiLayerDataTableBody.vue
fsoffel, line 1: As far as I see these changes are not part of the ticket description. What was your motivation?

dataland-frontend/src/components/forms/parts/fields/StringBaseDataPointFormField.vue
pioSko2: Does `:modelValue` work as two-way binding? Will it also update the value if it changes during Edit mode?

dataland-frontend/tests/e2e/fixtures/frameworks/gdv/GdvPreparedFixtures.ts
emanueldfine: Cannot set a property that is potentially undefined in TS.
Therefore I need to make sure the property exists before I can set it.

However it is implicitly expected that all of these properties are not null, because I use this function on a gdv-dataset without null values.

dataland-frontend/src/utils/DataSource.ts
pioSko2: Done.

dataland-frontend/src/utils/DataSource.ts
pioSko2: Done.

dataland-frontend/tests/component/utils/Constants.cy.ts
```diff
@@ -0,0 +1,23 @@
...
+    expect(sortedTestDataTypeEnums[0]).to.equal(DataTypeEnum.P2p);
+    expect(sortedTestDataTypeEnums[1]).to.equal(DataTypeEnum.Lksg);
+    expect(sortedTestDataTypeEnums[2]).to.equal(DataTypeEnum.Sme);
+    expect(sortedTestDataTypeEnums[3]).to.equal(DataTypeEnum.Gdv);
+  });
+
+  it("Check if the order stays the same if a list of data type enums does not contain gdv", () => {
+    const allDataTypeEnumsButGdv = Object.values(DataTypeEnum).filter((it) => it !== DataTypeEnum.Gdv);
+    const sortedAllDataTypeEnumsButGdv = putGdvAtTheEndOfList(allDataTypeEnumsButGdv);
+    expect(sortedAllDataTypeEnumsButGdv.length).to.equal(allDataTypeEnumsButGdv.length);
```
fsoffel, line 18: expect().to.deep.equal() would do it as well.

dataland-frontend/tests/component/components/forms/CreateGdvDataset.cy.ts
```diff
@@ -0,0 +1,23 @@
+import CreateGdvDataset from "@/components/forms/CreateGdvDataset.vue";
+import { minimalKeycloakMock } from "@ct/testUtils/Keycloak";
+import { submitButton } from "@sharedUtils/components/SubmitButton";
+
+describe("Component tests for the CreateP2pDataset that test dependent fields", () => {
```
fsoffel, line 5: Not P2p

dataland-frontend/tests/component/components/forms/CreateGdvDataset.cy.ts
```diff
@@ -0,0 +1,23 @@
...
+  it("On the upload page, ensure that sectors can be selected and deselected and the submit looks as expected", () => {
```
fsoffel, line 6: This is not about sectors

dataland-frontend/tests/e2e/fixtures/frameworks/gdv/GdvPreparedFixtures.ts
fsoffel: I see. Ok

dataland-framework-toolbox/src/main/kotlin/org/dataland/frameworktoolbox/intermediate/components/IntegerComponent.kt
fsoffel: 👍

dataland-framework-toolbox/src/main/kotlin/org/dataland/frameworktoolbox/intermediate/components/IntegerComponent.kt
fsoffel: ❤️

dataland-frontend/tests/e2e/fixtures/frameworks/gdv/GdvPreparedFixtures.ts
fsoffel: Cool

dataland-framework-toolbox/inputs/gdv/dataDictionary-GDV-VOEB.xlsx
fsoffel: Ok

dataland-frontend/src/utils/DataSource.ts
fsoffel: ❤️

dataland-frontend/tests/component/components/forms/CreateGdvDataset.cy.ts
emanueldfine: fixed

dataland-frontend/src/components/resources/dataTable/modals/GdvListOfBaseDataPointModal.vue
fsoffel: I find show-if makes more sense but it doesn't really matter. I'll stop wasting our time and close this.

dataland-frontend/src/components/forms/parts/fields/StringBaseDataPointFormField.vue
fsoffel: I think it is a one way binding. But this is fine since this field is hidden and thus can not manipulated by the user.

dataland-frontend/src/components/forms/parts/fields/StringBaseDataPointFormField.vue
fsoffel: That is good to know. I fear we have the same issue on other form elements. Should we open a ticket regarding that?

pioSko2: All inputs with documents have been checked and changed in DALA-3036. Other inputs seemed to work fine.
However, if you feel there may be other areas where we should check then sure. I'd be happy to test that, too.

pioSko2: This was specific for dataSource/documents

dataland-frontend/src/components/forms/parts/fields/StringBaseDataPointFormField.vue
fsoffel: Oh. On edit we need a two way binding actually I think.

dataland-frontend/src/components/forms/parts/fields/StringBaseDataPointFormField.vue
fsoffel: My bad. I somehow did not notice it at the other places, yet.

dataland-framework-toolbox/inputs/gdv/dataDictionary-GDV-VOEB-GDV-VÖB ESG questionnaire.csv
fsoffel, line 8: It is "Baugewerbe/Bau" but "Gewerbe / Herstellung". This feels inconsistent. Is this is intentional and the "/"s have different meanings. If you did not talk about that with Julia, I'd send her a message.

dataland-framework-toolbox/src/main/kotlin/org/dataland/frameworktoolbox/intermediate/components/IntegerComponent.kt
emanueldfine: Done

dataland-framework-toolbox/src/main/kotlin/org/dataland/frameworktoolbox/intermediate/components/IntegerComponent.kt
emanueldfine: DecimalComponent now also puts its unit into the uploadConfig entry.

dataland-framework-toolbox/inputs/gdv/dataDictionary-GDV-VOEB-GDV-VÖB ESG questionnaire.csv
emanueldfine: Added spaces to Baugwerbe / Bau

dataland-framework-toolbox/src/main/kotlin/org/dataland/frameworktoolbox/intermediate/components/basecomponents/NumberBaseComponent.kt
```diff
@@ -0,0 +1,49 @@
...
+
+/**
+ * The NumberBaseComponent serves as base-class for any number component (percentage, decimal, integer...) and
+ * offers some standard functionality.
+ */
+open class NumberBaseComponent(
+    identifier: String,
+    parent: FieldNodeParent,
+) : ComponentBase(identifier, parent) {
+    open var constantUnitSuffix: String? = null
```
fsoffel, line 21: Why did this ever needed to be open?

dataland-framework-toolbox/src/main/kotlin/org/dataland/frameworktoolbox/intermediate/components/basecomponents/NumberBaseComponent.kt
```diff
@@ -0,0 +1,49 @@
...
+
+    override fun generateDefaultDataModel(dataClassBuilder: DataClassBuilder) {
+        dataClassBuilder.addProperty(
+            this.identifier,
+            documentSupport.getJvmTypeReference(
+                TypeReference("java.math.BigDecimal", isNullable),
```
fsoffel, line 27: If the constructor had a parameter `fullyQualifiedName` and pass it to `TypeReference` instead of `java.math.BigDecimal` Then in no subclass be the need for overwriting this method because the subclasses could control it themselves.

dataland-framework-toolbox/src/main/kotlin/org/dataland/frameworktoolbox/intermediate/components/PercentageComponent.kt
```diff
@@ -41,6 +32,13 @@ class PercentageComponent(
         )
     }
 
+    override fun generateDefaultUploadConfig(uploadCategoryBuilder: UploadCategoryBuilder) {
+        uploadCategoryBuilder.addStandardUploadConfigCell(
+            component = this,
+            uploadComponentName = "PercentageFormField",
```
fsoffel, line 38: I think we are trying to move away from this upload component, because it is somewhat redundant. The job can be done by a `NumberFormField` with unit set to `%`. I think we did so already in some frameworks for percentage fields.

dataland-frontend/src/components/forms/parts/fields/StringBaseDataPointFormField.vue
pioSko2: Ahh yeah :) Done.

dataland-framework-toolbox/src/main/kotlin/org/dataland/frameworktoolbox/intermediate/components/PercentageComponent.kt
fsoffel: Also where was the upload config before? 😮

emanueldfine: Ok I can easily remove the PercentageComponent in this story, because there is no field anyways that uses it.
This only exists, because apparently Marc used it in the Heimathafen branch.
So should I delete it?  It wouldn't break anything in this branch here because it is unused.

The uploadconfig was not there before, because noone noticed that it is missing, because no field in GDV uses this component here.

dataland-framework-toolbox/src/main/kotlin/org/dataland/frameworktoolbox/intermediate/components/basecomponents/NumberBaseComponent.kt
emanueldfine: I can only imagine Marc doing this for good-practice.
In our case here, we don't need it as open var because none of the inheriting classes will override this anyways.

So should I make it a normal "var" ?

dataland-framework-toolbox/src/main/kotlin/org/dataland/frameworktoolbox/intermediate/components/basecomponents/NumberBaseComponent.kt
emanueldfine: I will investigate this idea and come back to this conversation afterwards.

dataland-frontend/src/components/forms/parts/fields/StringBaseDataPointFormField.vue
fsoffel: Why not remove the whole formkit tag?

pioSko2: hrmm.. wow didn't notice. Done.

dataland-frontend/src/components/forms/parts/fields/StringBaseDataPointFormField.vue
pioSko2: is this addressed or something needs to be changed?

dataland-framework-toolbox/src/main/kotlin/org/dataland/frameworktoolbox/frameworks/InDevelopmentPavedRoadFramework.kt
emanueldfine: In our call you said that you got this now. So can you close this pls?

dataland-frontend/src/components/forms/parts/fields/StringBaseDataPointFormField.vue
pioSko2: Oh, right. I didn't notice that. All other components with dataSource have the fileReference computed based on the fileName. So a `:modelValue` is sufficient there.
I'll change and test.

pioSko2: I changed it to v-model. In this case it seems more appropriate.

dataland-framework-toolbox/src/main/kotlin/org/dataland/frameworktoolbox/intermediate/components/basecomponents/NumberBaseComponent.kt
fsoffel: It has no use right now. But it doesn't bother me. So do whatever you want. Resolved.

dataland-frontend/src/components/resources/dataTable/MultiLayerDataTableBody.vue
fsoffel: TODO for me check this out for many datasets

dataland-frontend/src/components/resources/dataTable/MultiLayerDataTableBody.vue
fsoffel: TODO for me

dataland-framework-toolbox/src/main/kotlin/org/dataland/frameworktoolbox/template/components/ComponentFactoryContainer.kt
fsoffel: @MarcTM01

dataland-framework-toolbox/src/main/kotlin/org/dataland/frameworktoolbox/intermediate/components/basecomponents/NumberBaseComponent.kt
emanueldfine: I have pushed a change for this.   Can you check it and give me feedback ?

fsoffel: Love it.

dataland-frontend/src/components/forms/parts/fields/StringBaseDataPointFormField.vue
fsoffel: Ah. Now I get your point.

dataland-framework-toolbox/src/main/kotlin/org/dataland/frameworktoolbox/template/components/ComponentFactoryContainer.kt
MarcTM01: This wrapper is required to ensure that the list of factories is ordered according to the ordering annotations on the spring beans.

fsoffel: Ok. On https://www.baeldung.com/spring-order they do not proceed like that. But as it is working fine now, resolved.

dataland-framework-toolbox/inputs/gdv/dataDictionary-GDV-VOEB.xlsx
emanueldfine: It is gone

PR #1872, opened 2026-04-10
Dala 7163 sql query optimization

florian-hoeft:
# Pull Request \<Title>
`<Description here>`
  - [x] The changes have been inspected for possible breaking API changes (changed data models, endpoints, response codes, exception handling, ...). If there are any discuss them with the team.
  - [x] Make sure all newly added functionality (especially user facing) is tested
  - [x] Make sure that new test files are included in a test container and actually run in the CI
- [x] At least 3 consecutive CI runs are successfully executed to ensure that there are no flaky tests.
- [x] Documentation is updated as required. If unsure whether or what to update, ask a fellow developer.
- [x] ELSE, the new version is deployed to one of the dev servers using this branch
  - [x] The new feature is manually used/tested/observed on dev server. Testing of the new feature should (also) be done by a second, independent reviewer from the dev team
- [x] Confirm that the latest version (use the /gitinfo endpoint) of the feature branch is deployed to one of the dev servers (no longer enforced by GitHub)
- [x] Ensure that the release notes are written in a way that is understandable to a non-technical audience.


dataland-qa-service/src/main/kotlin/org/dataland/datalandqaservice/services/QaReviewQueryService.kt
```diff
@@ -90,8 +108,99 @@ class QaReviewQueryService
...
+            // Time adding priorities (calls external data sourcing service)
+            val addPrioritiesStartNs = System.nanoTime()
+            val withPriorities = addPrioritiesToResponse(qaReviewResponses)
+            logger.info(
+                "perf|getInfoOnPendingDatasets|addPrioritiesToResponse|datasetCount={} responseCount={} elapsedMs={}",
+                datasetIds.size,
+                withPriorities.size,
+                (System.nanoTime() - addPrioritiesStartNs) / NS_IN_MS,
+            )
+            return withPriorities
```
florian-hoeft: I think the name withPriorities could be improved. Can you think of a clearer name?

dataland-qa-service/src/main/kotlin/org/dataland/datalandqaservice/services/QaReviewQueryService.kt
```diff
@@ -90,8 +108,99 @@ class QaReviewQueryService
...
+                (System.nanoTime() - getJudgementsStartANs) / NS_IN_MS,
+            )
+
+            val getJudgementsStartBNs = System.nanoTime()
+
+            val latestJudgementByDatasetId =
+                judgementEntities
+                    .groupBy { it.datasetId }
+                    .mapValues { (_, judgements) ->
+                        val j = judgements.first()
```
florian-hoeft: If we care about the latest judgement here, should we add an explicit sort by timestamp?
This would mean adding an ORDER_BY (timestamp) in the repository.

florian-hoeft: Also, please use a more specific name than "j". Something like "firstJudgement" or "mostRecentJudgementForThisDatasetId" if we sort by timestamp. 
Those are just suggestions though.

dataland-qa-service/src/main/kotlin/org/dataland/datalandqaservice/services/QaReviewQueryService.kt
```diff
@@ -78,9 +90,15 @@ class QaReviewQueryService
          * which are still pending review (qaStatus = Pending).
          */
         @Transactional(readOnly = true)
-        fun getInfoOnPendingDatasets(companyName: String?): List<QaReviewResponse> {
+        fun getInfoOnPendingDatasets(
+            companyName: String?,
+        ): List<QaReviewResponse> {
+            val getInfoOnPendingDatasetsStartNs = System.nanoTime()
```
florian-hoeft: Please remove any logging of the runtime once you are done!

dataland-qa-service/src/main/kotlin/org/dataland/datalandqaservice/services/QaReviewQueryService.kt
```diff
@@ -90,8 +108,99 @@ class QaReviewQueryService
...
+
+            val datasetUUIDs = datasetIds.map { convertToUUID(it) }
+            // Time the dataset judgement fetch
+            val getJudgementsStartANs = System.nanoTime()
+            // Use a fetch-join query to load dataPoints together and avoid N+1
+            val judgementEntities =
+                try {
+                    datasetJudgementRepository.findAllByDatasetIdInWithDataPoints(datasetUUIDs)
+                } catch (ex: PersistenceException) {
+                    // If custom query fails for any reason related to persistence provider, fall back to original method
```
florian-hoeft: remove this comment

dataland-qa-service/src/main/kotlin/org/dataland/datalandqaservice/services/QaReviewQueryService.kt
```diff
@@ -90,8 +108,99 @@ class QaReviewQueryService
...
+                    logger.warn("Could not use fetch-join query for dataset judgements, falling back to default. Error: {}", ex.message)
+                    datasetJudgementRepository.findAllByDatasetIdIn(datasetUUIDs)
+                } catch (ex: DataAccessException) {
+                    // If Spring data access layer throws an error, fall back to original method
```
florian-hoeft: remove this comment

dataland-qa-service/src/main/kotlin/org/dataland/datalandqaservice/services/DataPointQaReportManager.kt
```diff
@@ -196,4 +196,15 @@ class DataPointQaReportManager(
      * @return number of matching QA reports
      */
     fun countQaReportsForDataPointIds(dataPointIds: Set<String>): Long = qaReportRepository.countByDataPointIdIn(dataPointIds)
+
+    /**
+     * Returns a map from dataPointId to the number of active QA reports for each data-point ID in the given set.
```
florian-hoeft: We usually write "data point" without the "-" in between. Could you make sure to correct that in the files you have touched?

dataland-qa-service/src/main/kotlin/org/dataland/datalandqaservice/services/QaReviewQueryService.kt
```diff
@@ -109,11 +218,34 @@ class QaReviewQueryService
                     DsBasicDataDimensions(it.companyId, it.framework, it.reportingPeriod)
                 }
 
+            // Time the external data sourcing call
+            val dataSourcingStartNs = System.nanoTime()
```
florian-hoeft: Please also remove any logging of the time here once you are done!

dataland-qa-service/src/main/kotlin/org/dataland/datalandqaservice/services/QaReviewQueryService.kt
```diff
@@ -17,13 +17,17 @@ import org.dataland.datalandqaservice.repositories.QaReviewRepository
...
+import org.springframework.dao.DataAccessException
+import jakarta.persistence.PersistenceException
 import org.springframework.beans.factory.annotation.Autowired
 import org.springframework.http.HttpStatus
 import org.springframework.stereotype.Service
 import org.springframework.transaction.annotation.Transactional
 import java.util.UUID
 import org.dataland.dataSourcingService.openApiClient.model.BasicDataDimensions as DsBasicDataDimensions
 
+const val NS_IN_MS = 1_000_000
```
florian-hoeft: remove once you are done with checking and logging runtime

dataland-qa-service/src/main/kotlin/org/dataland/datalandqaservice/services/QaReviewQueryService.kt
```diff
@@ -222,27 +354,111 @@ class QaReviewQueryService
...
+                allDataPointIds.size,
+                (System.nanoTime() - getNumberOfQaReportsForDatasetIdsStartNs) / NS_IN_MS,
+            )
+            return qaReportsByDatasetId
+        }
+
+        /**
+         * Converts the QaReviewEntity into a QaReviewResponse using pre-fetched data to avoid per-item I/O.
+         */
+        private fun QaReviewEntity.toQaReviewResponse(
```
florian-hoeft: Could you rename this overload? I think it is confusing since you now have two ".toQaReviewResponse" overloads. Please rename in such a way that makes it clear these are two distinct calls. 
One suggestion would be ".toQaReviewResponseWithPrecomputedData"

dataland-qa-service/src/main/kotlin/org/dataland/datalandqaservice/services/QaReviewQueryService.kt
```diff
@@ -90,8 +108,99 @@ class QaReviewQueryService
...
+                "perf|getInfoOnPendingDatasets|getNumberOfQaReportsForDatasetIds|datasetCount={} elapsedMs={}",
+                datasetIds.size,
+                (System.nanoTime() - getNumberOfQaReportsForDatasetIdsStartNs) / NS_IN_MS,
+            )
+
+            val datasetUUIDs = datasetIds.map { convertToUUID(it) }
+            // Time the dataset judgement fetch
+            val getJudgementsStartANs = System.nanoTime()
+            // Use a fetch-join query to load dataPoints together and avoid N+1
+            val judgementEntities =
```
florian-hoeft: I like the error handling here. Good job. Could you also log the class of the exception with the message?

dataland-qa-service/src/main/kotlin/org/dataland/datalandqaservice/services/QaReviewQueryService.kt
```diff
@@ -78,9 +90,15 @@ class QaReviewQueryService
          * which are still pending review (qaStatus = Pending).
          */
         @Transactional(readOnly = true)
-        fun getInfoOnPendingDatasets(companyName: String?): List<QaReviewResponse> {
+        fun getInfoOnPendingDatasets(
```
florian-hoeft: We definitely want to add/confirm an integration test that checks getInfoOnPendingDatasets works as intended and returns the same List as before. 
This comment is just a reminder for myself to check this later.

dataland-qa-service/src/main/kotlin/org/dataland/datalandqaservice/services/QaReviewQueryService.kt
```diff
@@ -222,27 +354,111 @@ class QaReviewQueryService
...
                 }
             }
 
+        /**
+         * Returns a map from datasetId to the number of QA reports for all data points contained in that dataset.
+         * Fetches metadata for all datasetIds in bulk (one call per datasetId to the metadata API) and counts
+         * QA reports for all collected data-point IDs in a single DB query.
+         */
+        private fun getNumberOfQaReportsForDatasetIds(datasetIds: List<String>): Map<String, Long> {
+            val getNumberOfQaReportsForDatasetIdsStartNs = System.nanoTime()
```
florian-hoeft: Reminder to remove the logging of runtime here as well.

dataland-qa-service/src/main/kotlin/org/dataland/datalandqaservice/services/QaReviewQueryService.kt
```diff
@@ -222,27 +354,111 @@ class QaReviewQueryService
...
+                            (System.nanoTime() - getContainedDataPointsStartNs) / NS_IN_MS,
+                        )
+                        dataPointIdsByDatasetId[datasetId] = emptySet()
+                    } else {
+                        throw clientException
+                    }
+                }
+            }
+            val allDataPointIds = dataPointIdsByDatasetId.values.flatten().toSet()
+            val countQaReportsForDataPointIdsBulkStartNs = System.nanoTime()
```
florian-hoeft: reminder to remove this runtime logging

dataland-qa-service/src/main/kotlin/org/dataland/datalandqaservice/repositories/DatasetJudgementRepository.kt
```diff
@@ -17,4 +19,22 @@ interface DatasetJudgementRepository : JpaRepository<DatasetJudgementEntity, UUI
...
+     * Finds all DatasetJudgementEntity objects whose datasetId is in the given collection.
+     * Used for bulk (batch) lookups to avoid N individual queries.
+     *
+     * @param datasetIds The collection of dataset IDs to look up.
+     * @return A list of DatasetJudgementEntity objects associated with any of the given datasetIds.
+     */
+    fun findAllByDatasetIdIn(datasetIds: Collection<UUID>): List<DatasetJudgementEntity>
+
+    /**
+     * Fetches dataset judgements together with their associated dataPoints in a single query to avoid
```
florian-hoeft: please write "data points" instead of dataPoints here and elsewhere.

dataland-qa-service/src/main/kotlin/org/dataland/datalandqaservice/repositories/DatasetJudgementRepository.kt
```diff
@@ -17,4 +19,22 @@ interface DatasetJudgementRepository : JpaRepository<DatasetJudgementEntity, UUI
...
+     * N+1 queries when converting entities to response objects that touch the collection.
+     */
+    @Query("select distinct d from DatasetJudgementEntity d left join fetch d.dataPoints where d.datasetId in :datasetIds")
+    fun findAllByDatasetIdInWithDataPoints(
```
florian-hoeft: Since the query is explicit here, I would rename this to something like "findAllWithDataPointsByDatasetIdIn"

dataland-qa-service/src/main/kotlin/org/dataland/datalandqaservice/repositories/DataPointQaReportRepository.kt
```diff
@@ -61,6 +61,22 @@ interface DataPointQaReportRepository : JpaRepository<DataPointQaReportEntity, S
         @Param("dataPointIds") dataPointIds: Set<String>,
     ): Long
 
+    /**
+     * Returns per-data-point active QA report counts for all data-point IDs in the given set.
```
florian-hoeft: "data point" instead of "data-point" please

florian-hoeft (commented):
First round review

dataland-qa-service/src/main/kotlin/org/dataland/datalandqaservice/services/QaReviewQueryService.kt
```diff
@@ -90,8 +108,99 @@ class QaReviewQueryService
...
-                    ).map { it.toQaReviewResponse(userIsAdmin) }
-            return addPrioritiesToResponse(qaReviewResponses)
+                    )
+            logger.info(
+                "perf|getInfoOnPendingDatasets|getPendingQaReviewMetadatasetsByCompany|datasetCount={} elapsedMs={}",
+                entities.size,
+                (System.nanoTime() - getPendingStartNs) / NS_IN_MS,
+            )
+
+            val datasetIds = entities.map { it.dataId }
```
florian-hoeft: Possibly do val datasetIds = entities.map { it.dataId }.distinct() 
if there can be multiple QaReviewEntities pending per dataId. I dont think this is the case, but it would not hurt to double check this.

dataland-qa-service/src/main/kotlin/org/dataland/datalandqaservice/repositories/DataPointQaReportRepository.kt
```diff
@@ -61,6 +61,22 @@ interface DataPointQaReportRepository : JpaRepository<DataPointQaReportEntity, S
...
+     */
+    @Query(
+        "SELECT qaReport.dataPointId, COUNT(qaReport) FROM DataPointQaReportEntity qaReport " +
+            "WHERE qaReport.dataPointId IN :dataPointIds " +
+            "AND qaReport.active = TRUE " +
+            "GROUP BY qaReport.dataPointId",
+    )
+    fun countByDataPointIdInGrouped(
+        @Param("dataPointIds") dataPointIds: Set<String>,
+    ): List<Array<Any>>
```
florian-hoeft: I think returning List<Array<Any>> could be brittle here. Especially since we do ".associate { row -> row[0] as String to row[1] as Long }"
with the return. 

I would suggest introducing an interface:
interface DataPointIdCountProjection {
val dataPointId: String
val activeQaReportCount: Long
}

And then returning: List<DataPointIdCountProjection>. 

And then 

fun countQaReportsForDataPointIdsBulk(dataPointIds: Set<String>): Map<String, Long> =
qaReportRepository
.countByDataPointIdInGrouped(dataPointIds)
.associate { it.dataPointId to it.cnt }

florian-hoeft (commented):
First round review part 2/2

dataland-qa-service/src/main/kotlin/org/dataland/datalandqaservice/services/QaReviewQueryService.kt
ivan-gila: ok, done

dataland-qa-service/src/main/kotlin/org/dataland/datalandqaservice/services/QaReviewQueryService.kt
ivan-gila: ok

dataland-qa-service/src/main/kotlin/org/dataland/datalandqaservice/services/QaReviewQueryService.kt
ivan-gila: done

dataland-qa-service/src/main/kotlin/org/dataland/datalandqaservice/services/DataPointQaReportManager.kt
ivan-gila: yes, done

dataland-qa-service/src/main/kotlin/org/dataland/datalandqaservice/services/QaReviewQueryService.kt
ivan-gila: done

dataland-qa-service/src/main/kotlin/org/dataland/datalandqaservice/services/QaReviewQueryService.kt
ivan-gila: yes, qaReviewResponsesWithPriorities looks better

dataland-qa-service/src/main/kotlin/org/dataland/datalandqaservice/services/QaReviewQueryService.kt
ivan-gila: done

dataland-qa-service/src/main/kotlin/org/dataland/datalandqaservice/repositories/DatasetJudgementRepository.kt
ivan-gila: done

dataland-qa-service/src/main/kotlin/org/dataland/datalandqaservice/repositories/DatasetJudgementRepository.kt
ivan-gila: ok

dataland-qa-service/src/main/kotlin/org/dataland/datalandqaservice/repositories/DataPointQaReportRepository.kt
ivan-gila: ok

dataland-qa-service/src/main/kotlin/org/dataland/datalandqaservice/services/QaReviewQueryService.kt
ivan-gila: is an ORDER_BY timestamp in the query enough to guarantee the judgements to be time-ordered?

dataland-qa-service/src/main/kotlin/org/dataland/datalandqaservice/repositories/DataPointQaReportRepository.kt
ivan-gila: done, I've implemented the new interface in the DataPointIdCountProjection.kt file

dataland-qa-service/src/main/kotlin/org/dataland/datalandqaservice/services/QaReviewQueryService.kt
florian-hoeft: Good. Will close this conversation.

dataland-qa-service/src/main/kotlin/org/dataland/datalandqaservice/services/QaReviewQueryService.kt
florian-hoeft: To give more context to this comment: I wrote it because the "val j = judgements.first()" line assumes that what is coming from the backend is ordered by creation time. 
I do not know if this is always the case, though it seems to be. 

Thus, to have a fallback here, we could do "val j = sortedByDescending {it.timestamp}" to be more certain. 

Then, I do not think an ORDER_BY clause in the query is necessary.

dataland-qa-service/src/main/kotlin/org/dataland/datalandqaservice/services/QaReviewQueryService.kt
florian-hoeft: Thanks. Will close the conversation.

dataland-qa-service/src/main/kotlin/org/dataland/datalandqaservice/services/QaReviewQueryService.kt
florian-hoeft: Thanks. Will close this conversation.

dataland-qa-service/src/main/kotlin/org/dataland/datalandqaservice/services/QaReviewQueryService.kt
ivan-gila: this substitution causes an error, it seems because DatasetJudgementEntity does not have a timestamp field, do I add this field to it?

dataland-qa-service/src/main/kotlin/org/dataland/datalandqaservice/services/DataPointQaReportManager.kt
florian-hoeft: Thank you. Will close this conversation.

dataland-qa-service/src/main/kotlin/org/dataland/datalandqaservice/services/QaReviewQueryService.kt
ivan-gila: I've added a timestamp attribute to datasetJudgementEntity in DatasetJudgementCreationService.kt:

```timestamp = datasetMetaData.uploadTime```

Comment (florian-hoeft) [reply dataland-qa-service/src/main/kotlin/org/dataland/datalandqaservice/services/QaReviewQueryService.kt]: True. I will check the backend and come back to you on this issue. For now, you can leave it as you had before.

Comment (ivan-gila) [reply dataland-qa-service/src/main/kotlin/org/dataland/datalandqaservice/services/QaReviewQueryService.kt]: ok, so I'll wait for commit/push my modifications on this

Comment (florian-hoeft) [reply dataland-qa-service/src/main/kotlin/org/dataland/datalandqaservice/services/QaReviewQueryService.kt]: Okay, I checked the backend: DatasetJudgementEntities are stored to the DB one after the other. However, that does not guarantee absolute order by creation time, because postgreSQL may sometimes change order for various events. 

I will create a seperate ticket regarding this, so we can move on with this branch.

File: dataland-qa-service/src/main/kotlin/org/dataland/datalandqaservice/services/QaReviewQueryService.kt (hunk 10 above)
Comment (florian-hoeft) [reply dataland-qa-service/src/main/kotlin/org/dataland/datalandqaservice/services/QaReviewQueryService.kt]: Good, but now my IDE is telling me the lines are too long? Are you using prettier (the code formatter) on safe?

File: dataland-qa-service/src/main/kotlin/org/dataland/datalandqaservice/services/QaReviewQueryService.kt (hunk 2 above)
Comment (ivan-gila) [reply dataland-qa-service/src/main/kotlin/org/dataland/datalandqaservice/services/QaReviewQueryService.kt]: ok, as you prefer (the timestamp that I've added should be sufficient, I've not committed it yet)

File: dataland-qa-service/src/main/kotlin/org/dataland/datalandqaservice/services/QaReviewQueryService.kt (hunk 10 above)
Comment (ivan-gila) [reply dataland-qa-service/src/main/kotlin/org/dataland/datalandqaservice/services/QaReviewQueryService.kt]: fixed

File: dataland-qa-service/src/main/kotlin/org/dataland/datalandqaservice/services/QaReviewQueryService.kt (hunk 2 above)
Comment (florian-hoeft) [reply dataland-qa-service/src/main/kotlin/org/dataland/datalandqaservice/services/QaReviewQueryService.kt]: Please revert your changes regarding the timestamp. 
What I addressed in this conversation will be part of a seperate ticket.

Comment (ivan-gila) [reply dataland-qa-service/src/main/kotlin/org/dataland/datalandqaservice/services/QaReviewQueryService.kt]: ok, done

Comment (ivan-gila) [reply dataland-qa-service/src/main/kotlin/org/dataland/datalandqaservice/services/QaReviewQueryService.kt]: done

Comment (florian-hoeft) [reply dataland-qa-service/src/main/kotlin/org/dataland/datalandqaservice/services/QaReviewQueryService.kt]: I created the follow up ticket and will close this conversation.

File: dataland-qa-service/src/main/kotlin/org/dataland/datalandqaservice/services/QaReviewQueryService.kt (hunk 10 above)
Comment (florian-hoeft) [reply dataland-qa-service/src/main/kotlin/org/dataland/datalandqaservice/services/QaReviewQueryService.kt]: Thanks. The two logger.warn messages are practically identical. Could you group them in one catch() block to reduce boilerplate?

File: dataland-qa-service/src/main/kotlin/org/dataland/datalandqaservice/services/QaReviewQueryService.kt (hunk 2 above)
Comment (florian-hoeft) [reply dataland-qa-service/src/main/kotlin/org/dataland/datalandqaservice/services/QaReviewQueryService.kt]: Perfect. Conversation closed.

File: dataland-qa-service/src/main/kotlin/org/dataland/datalandqaservice/services/QaReviewQueryService.kt (hunk 9 above)
Comment (florian-hoeft) [reply dataland-qa-service/src/main/kotlin/org/dataland/datalandqaservice/services/QaReviewQueryService.kt]: Good. Resolved.

File: dataland-qa-service/src/main/kotlin/org/dataland/datalandqaservice/repositories/DatasetJudgementRepository.kt (hunk 14 above)
Comment (florian-hoeft) [reply dataland-qa-service/src/main/kotlin/org/dataland/datalandqaservice/repositories/DatasetJudgementRepository.kt]: Thanks. Resolved.

File: dataland-qa-service/src/main/kotlin/org/dataland/datalandqaservice/repositories/DatasetJudgementRepository.kt (hunk 15 above)
Comment (florian-hoeft) [reply dataland-qa-service/src/main/kotlin/org/dataland/datalandqaservice/repositories/DatasetJudgementRepository.kt]: Resolved.

File: dataland-qa-service/src/main/kotlin/org/dataland/datalandqaservice/repositories/DataPointQaReportRepository.kt (hunk 18 above)
Comment (florian-hoeft) [reply dataland-qa-service/src/main/kotlin/org/dataland/datalandqaservice/repositories/DataPointQaReportRepository.kt]: Perfect. Resolved.

File: dataland-qa-service/src/main/kotlin/org/dataland/datalandqaservice/services/QaReviewQueryService.kt (hunk 10 above)
Comment (ivan-gila) [reply dataland-qa-service/src/main/kotlin/org/dataland/datalandqaservice/services/QaReviewQueryService.kt]: done

File: dataland-qa-service/src/main/kotlin/org/dataland/datalandqaservice/repositories/DataPointIdCountProjection.kt [hunk 19]
```diff
@@ -0,0 +1,10 @@
+package org.dataland.datalandqaservice.org.dataland.datalandqaservice.repositories
```
florian-hoeft: I do not think this needs to be a repository (in the repositories directory), since it is a simple interface. 
Can you double check?  
To me, it seems this could be directly in "DataPointQaReportRepository" as a small helper interface or in the "utils" directory

ivan-gila: I agree, I've inserted it into DataPointQaReportRepository

dataland-qa-service/src/main/kotlin/org/dataland/datalandqaservice/services/QaReviewQueryService.kt
```diff
@@ -41,6 +45,14 @@ class QaReviewQueryService
     ) {
         private val logger = LoggerFactory.getLogger(javaClass)
 
+        // Lightweight container used to avoid building full DatasetJudgementResponse
+        // which accesses collections (and can trigger extra queries).
+        private data class MinimalDatasetJudgement(
+            val dataSetJudgementId: String,
+            val qaJudgeUserId: String?,
```
florian-hoeft: Why are the qaJudgeUserId and qaJudgeUserName optional here?

ivan-gila: I just required the dataSetJudgementId to be mandatory, not sure it was the right thing

Now I've required also qaJudgeUserId and qaJudgeUserName to be mandatory

dataland-qa-service/src/main/kotlin/org/dataland/datalandqaservice/services/QaReviewQueryService.kt
ivan-gila: wait, the solution I've implemented (reported below) leads to a gradle error `TooGenericExceptionCaught - [The caught exception is too generic. Prefer catching specific exceptions to the (...)] at /home/runner/work/Dataland/Dataland/dataland-qa-service/src/main/kotlin/org/dataland/datalandqaservice/services/QaReviewQueryService.kt:133:26`

solution:

```
    try {
        datasetJudgementRepository.findAllWithDataPointsByDatasetIdIn(datasetUUIDs)
    } catch (ex: Exception) {
        if (ex is PersistenceException || ex is DataAccessException) {
            logger.warn(
                "Could not use fetch-join query for dataset judgements, falling back to default. Error [{}]: {}",
                ex::class.simpleName,
                ex.message,
            )
            datasetJudgementRepository.findAllByDatasetIdIn(datasetUUIDs)
        } else {
            throw ex
        }
    }
                
```

I've kept the two catch separated and implemented a single function called in both

ivan-gila: Kotlin does not admit the syntax `catch (ex is PersistenceException || ex is DataAccessException)` and `catch (ex: Exception)` is too generic

dataland-qa-service/src/test/kotlin/org/dataland/datalandqaservice/services/QaReviewQueryServiceTest.kt
```diff
@@ -190,4 +191,146 @@ class QaReviewQueryServiceTest {
         Assertions.assertEquals(1, responses.size)
         Assertions.assertNull(responses.first().priorityOfAssociatedDataSourcing)
     }
+
+    @Test
```
florian-hoeft, line 269: Regarding the tests, I suggest adding coverage for two additional scenarios:

1) Fallback path for dataset judgements:
    - Add a test where `datasetJudgementRepository.findAllWithDataPointsByDatasetIdIn(...)` throws a `PersistenceException` (or `DataAccessException`).
    - Verify that `getInfoOnPendingDatasets` then calls `findAllByDatasetIdIn(...)` instead and still returns the correct judge info in the responses.

2) Dataset without any judgement:
    - Add a test where `getPendingQaReviewMetadatasetsByCompany(...)` returns one `QaReviewEntity`, and `findAllWithDataPointsByDatasetIdIn(...)` returns an empty list.
    - Verify that `getInfoOnPendingDatasets` still returns that dataset, but with `qaJudgeUserId`, `qaJudgeUserName`, and `datasetReviewId` all `null`, and with the correct `numberQaReports` / priorities.

dataland-qa-service/src/test/kotlin/org/dataland/datalandqaservice/services/QaReviewQueryServiceTest.kt
```diff
@@ -190,4 +287,50 @@ class QaReviewQueryServiceTest {
         Assertions.assertEquals(1, responses.size)
         Assertions.assertNull(responses.first().priorityOfAssociatedDataSourcing)
     }
+
```
florian-hoeft, line 268: Also can you explicitly stub the following calls:  metaDataControllerApi.getContainedDataPoints(...); 
dataPointQaReportManager.countQaReportsForDataPointIdsBulk(...); 
datasetJudgementRepository.findAllWithDataPointsByDatasetIdIn(...)

in setup(). In the old tests above where getInfoOnPendingDatasets is called (check that getInfoOnPendingDatasets fetches data sourcing priority of null if it is not specified, check that getInfoOnPendingDatasets fetches correct data sourcing priority), these calls are made but not stubbed in the code, which can be confusing.

dataland-qa-service/src/main/kotlin/org/dataland/datalandqaservice/services/QaReviewQueryService.kt
ivan-gila: yes, done

dataland-qa-service/src/main/kotlin/org/dataland/datalandqaservice/services/QaReviewQueryService.kt
ivan-gila: removed

dataland-qa-service/src/main/kotlin/org/dataland/datalandqaservice/services/QaReviewQueryService.kt
ivan-gila: removed

dataland-qa-service/src/main/kotlin/org/dataland/datalandqaservice/services/QaReviewQueryService.kt
ivan-gila: removed

dataland-qa-service/src/main/kotlin/org/dataland/datalandqaservice/services/QaReviewQueryService.kt
ivan-gila: removed

dataland-qa-service/src/test/kotlin/org/dataland/datalandqaservice/services/QaReviewQueryServiceTest.kt
ivan-gila: Tests implemented.

dataland-qa-service/src/test/kotlin/org/dataland/datalandqaservice/services/QaReviewQueryServiceTest.kt
ivan-gila: done

dataland-qa-service/src/main/kotlin/org/dataland/datalandqaservice/repositories/DatasetJudgementRepository.kt
```diff
@@ -17,4 +19,22 @@ interface DatasetJudgementRepository : JpaRepository<DatasetJudgementEntity, UUI
...
+     * @param datasetIds The collection of dataset IDs to look up.
+     * @return A list of DatasetJudgementEntity objects associated with any of the given datasetIds.
+     */
+    fun findAllByDatasetIdIn(datasetIds: Collection<UUID>): List<DatasetJudgementEntity>
+
+    /**
+     * Fetches dataset judgements together with their associated data points in a single query to avoid
+     * N+1 queries when converting entities to response objects that touch the collection.
+     */
+    @Query("select distinct d from DatasetJudgementEntity d left join fetch d.dataPoints where d.datasetId in :datasetIds")
```
SiegfriedSobkowiak: use common formatting here

dataland-qa-service/src/main/kotlin/org/dataland/datalandqaservice/services/QaReviewQueryService.kt
```diff
@@ -90,8 +100,54 @@ class QaReviewQueryService
...
+                    "Could not use fetch-join query for dataset judgements, falling back to default. Error [{}]: {}",
+                    ex::class.simpleName,
+                    ex.message,
+                )
+                datasetJudgementRepository.findAllByDatasetIdIn(datasetUUIDs)
+            }
+
+            val datasetUUIDs = datasetIds.map { convertToUUID(it) }
+
+            val judgementEntities =
```
florian-hoeft: Integrate a check that judgementEntities has only one entry and throw and error message if it does not

dataland-qa-service/src/main/kotlin/org/dataland/datalandqaservice/repositories/DatasetJudgementRepository.kt
ivan-gila: done

dataland-qa-service/src/main/kotlin/org/dataland/datalandqaservice/services/QaReviewQueryService.kt
```diff
@@ -90,8 +100,54 @@ class QaReviewQueryService
                             companyName = companyName,
                             qaStatuses = setOf(QaStatus.Pending),
                         ),
-                    ).map { it.toQaReviewResponse(userIsAdmin) }
-            return addPrioritiesToResponse(qaReviewResponses)
+                    )
+            val datasetIds = entities.map { it.dataId }.distinct()
+            val numberQaReportsByDatasetId = getNumberOfQaReportsForDatasetIds(datasetIds)
+
+            fun fallbackToNonFetch(
```
florian-hoeft: make this a private function and move it out of getInfoOnPendingDatasets

ivan-gila: done

dataland-qa-service/src/main/kotlin/org/dataland/datalandqaservice/repositories/DataPointQaReportRepository.kt
ivan-gila: Improved following suggestion of Siegfried

dataland-qa-service/src/main/kotlin/org/dataland/datalandqaservice/services/QaReviewQueryService.kt
ivan-gila: done

dataland-qa-service/src/main/kotlin/org/dataland/datalandqaservice/services/QaReviewQueryService.kt
florian-hoeft: Confirmed and resolved.

dataland-qa-service/src/main/kotlin/org/dataland/datalandqaservice/services/QaReviewQueryService.kt
florian-hoeft: Confirmed and resolved.

dataland-qa-service/src/main/kotlin/org/dataland/datalandqaservice/services/QaReviewQueryService.kt
florian-hoeft: Confirmed and resolved.

dataland-qa-service/src/main/kotlin/org/dataland/datalandqaservice/services/QaReviewQueryService.kt
florian-hoeft: Reminder to move the "fallbackToNonFetch" function out of the scope of the getInfoOnPendingDatasets function and make it a private fun

dataland-qa-service/src/main/kotlin/org/dataland/datalandqaservice/services/QaReviewQueryService.kt
florian-hoeft: Confirmed and resolved.

dataland-qa-service/src/main/kotlin/org/dataland/datalandqaservice/services/QaReviewQueryService.kt
florian-hoeft: confirmed and resolved.

dataland-qa-service/src/main/kotlin/org/dataland/datalandqaservice/repositories/DataPointIdCountProjection.kt
florian-hoeft: As Siegfried proposed, the new interfact should follow the scheme: 
_interface KeyValuePair {
      fun getK(): String
      fun getV(): String
  }
 
  @Query("SELECT e.key AS k, e.value AS v FROM MyEntity e")
  fun findAll(): List<KeyValuePair>
  
  With a map: 
  val map = repo.findAll().associate { it.getK() to it.getV() }_
  
  
  Please check if this is a valid way to do it. If so, implement it. If not, come back to me.

dataland-qa-service/src/main/kotlin/org/dataland/datalandqaservice/services/QaReviewQueryService.kt
florian-hoeft: Good. I checked the backend implementation and both fields are always stored in a datasetReviewEntity. Thus, they do not have to be implemented manually. I will close this conversation.

dataland-qa-service/src/test/kotlin/org/dataland/datalandqaservice/services/QaReviewQueryServiceTest.kt
florian-hoeft: Good. Resolved.

dataland-qa-service/src/main/kotlin/org/dataland/datalandqaservice/services/QaReviewQueryService.kt
ivan-gila: done

dataland-qa-service/src/main/kotlin/org/dataland/datalandqaservice/repositories/DataPointIdCountProjection.kt
ivan-gila: done

dataland-qa-service/src/main/kotlin/org/dataland/datalandqaservice/services/QaReviewQueryService.kt
```diff
@@ -113,12 +165,28 @@ class QaReviewQueryService
...
+                        null
+                    } else {
+                        throw ex
+                    }
                 }
 
             return QaReviewUtils.assignPriorities(qaReviewResponses, prioritiesOfAssociatedDataSourcing)
         }
 
+        private fun fallbackToNonFetch(
```
florian-hoeft: Please add KDoc (documentation)

dataland-qa-service/src/main/kotlin/org/dataland/datalandqaservice/repositories/DataPointQaReportRepository.kt
```diff
@@ -71,3 +87,18 @@ interface DataPointQaReportRepository : JpaRepository<DataPointQaReportEntity, S
     )
     fun findDataPointTypeUsingId(qaReportId: String): String
 }
+
+/**
+ * Projection interface for data point id / count pairs returned by queries.
+ */
+interface DataPointCount {
+    /**
+     * Return dataPointId
```
florian-hoeft: I do not think this comment is necessary.

dataland-qa-service/src/main/kotlin/org/dataland/datalandqaservice/repositories/DataPointQaReportRepository.kt
```diff
@@ -71,3 +87,18 @@ interface DataPointQaReportRepository : JpaRepository<DataPointQaReportEntity, S
...
+     */
+    fun getDataPointId(): String
+
+    /**
+     * Return count
```
florian-hoeft: I also do not think this comment is necessary.

dataland-qa-service/src/main/kotlin/org/dataland/datalandqaservice/repositories/DataPointQaReportRepository.kt
```diff
@@ -61,6 +61,22 @@ interface DataPointQaReportRepository : JpaRepository<DataPointQaReportEntity, S
         @Param("dataPointIds") dataPointIds: Set<String>,
     ): Long
 
+    /**
+     * Returns per data point active QA report counts for all data point IDs in the given set.
+     * Each element of the returned list is an array [dataPointId, count].
```
florian-hoeft: Please update the KDoc here.

dataland-qa-service/src/main/kotlin/org/dataland/datalandqaservice/repositories/DataPointQaReportRepository.kt
```diff
@@ -71,3 +87,18 @@ interface DataPointQaReportRepository : JpaRepository<DataPointQaReportEntity, S
...
+interface DataPointCount {
+    /**
+     * Return dataPointId
+     */
+    fun getDataPointId(): String
+
+    /**
+     * Return count
+     */
+    fun getCount(): Long
```
florian-hoeft: Also, please rename this to getQaReportCount() and the associated entry in the query from count to qaReportCount. This is because COUNT is also a keyword in SQL and we want to avoid using that as the alias and method name

dataland-qa-service/src/main/kotlin/org/dataland/datalandqaservice/services/DataPointQaReportManager.kt
```diff
@@ -196,4 +196,15 @@ class DataPointQaReportManager(
...
      */
     fun countQaReportsForDataPointIds(dataPointIds: Set<String>): Long = qaReportRepository.countByDataPointIdIn(dataPointIds)
+
+    /**
+     * Returns a map from dataPointId to the number of active QA reports for each data point ID in the given set.
+     * Uses a single grouped DB query instead of one query per data point.
+     * @param dataPointIds set of dataPointId values to count QA reports for
+     * @return map of dataPointId to count
+     */
+    fun countQaReportsForDataPointIdsBulk(dataPointIds: Set<String>): Map<String, Long> =
```
florian-hoeft: Could you add a small unit/integration test for this function?

dataland-qa-service/src/test/kotlin/org/dataland/datalandqaservice/services/QaReviewQueryServiceTest.kt
ivan-gila: Regarding the creation of objects for testing, I've implemented the option recommended by Siegfried (write a private fun that builds the required objects) in the same file `QaReviewQueryServiceTest.kt`. However, Sonar then revealed some code smells due to the file length (longer than 400 lines), so I moved the definition of this object into a new file named `QaReviewTestFixtures.kt`.

dataland-qa-service/src/main/kotlin/org/dataland/datalandqaservice/repositories/DataPointIdCountProjection.kt
florian-hoeft: Thanks. Resolved.

dataland-qa-service/src/main/kotlin/org/dataland/datalandqaservice/services/QaReviewQueryService.kt
florian-hoeft: Thanks. Could you add some KDoc to the function "fallbackToNonFetch"?

dataland-qa-service/src/test/kotlin/org/dataland/datalandqaservice/services/QaReviewTestFixtures.kt
```diff
@@ -0,0 +1,112 @@
...
+            triggeringUserId = args.triggeringUserId,
+            comment = args.comment,
+        )
+
+    val datasetJudgement =
+        DatasetJudgementEntity(
+            dataSetJudgementId = args.judgementId,
+            datasetId = UUID.fromString(args.dataId),
+            companyId = UUID.randomUUID(),
+            dataType = DataTypeEnum.sfdr,
```
florian-hoeft: You allow framework to vary when building the qaReview, but here you hardcode it to sfdr. Please derive it from "framework".

dataland-qa-service/src/test/kotlin/org/dataland/datalandqaservice/services/QaReviewTestFixtures.kt
```diff
@@ -0,0 +1,112 @@
+package org.dataland.datalandqaservice.services
```
florian-hoeft: Please move the file from the "services" directory to the "utils" directory

florian-hoeft (commented):
2nd round review comments

dataland-qa-service/src/main/kotlin/org/dataland/datalandqaservice/services/QaReviewQueryService.kt
ivan-gila: done

dataland-qa-service/src/main/kotlin/org/dataland/datalandqaservice/services/QaReviewQueryService.kt
ivan-gila: done

dataland-qa-service/src/main/kotlin/org/dataland/datalandqaservice/repositories/DataPointQaReportRepository.kt
ivan-gila: I put because otherwise ktlint returns an error: missing kdoc

dataland-qa-service/src/main/kotlin/org/dataland/datalandqaservice/repositories/DataPointQaReportRepository.kt
ivan-gila: I put because otherwise ktlint returns an error: missing kdoc

dataland-qa-service/src/main/kotlin/org/dataland/datalandqaservice/repositories/DataPointQaReportRepository.kt
ivan-gila: done

dataland-qa-service/src/main/kotlin/org/dataland/datalandqaservice/repositories/DataPointQaReportRepository.kt
ivan-gila: done

dataland-qa-service/src/test/kotlin/org/dataland/datalandqaservice/services/QaReviewTestFixtures.kt
ivan-gila: agree, done

dataland-qa-service/src/test/kotlin/org/dataland/datalandqaservice/services/QaReviewTestFixtures.kt
ivan-gila: done

dataland-qa-service/src/main/kotlin/org/dataland/datalandqaservice/services/QaReviewQueryService.kt
florian-hoeft: Thanks. Resolved.

dataland-qa-service/src/test/kotlin/org/dataland/datalandqaservice/services/QaReviewQueryServiceTest.kt
florian-hoeft: Yes that is valid thank you. I will close this conversation for now.

dataland-qa-service/src/main/kotlin/org/dataland/datalandqaservice/repositories/DatasetJudgementRepository.kt
florian-hoeft: Resolved.

dataland-qa-service/src/main/kotlin/org/dataland/datalandqaservice/services/QaReviewQueryService.kt
florian-hoeft: Resolved.

dataland-qa-service/src/main/kotlin/org/dataland/datalandqaservice/services/QaReviewQueryService.kt
florian-hoeft: Thanks. Resolved.

dataland-qa-service/src/main/kotlin/org/dataland/datalandqaservice/services/QaReviewQueryService.kt
florian-hoeft: Thanks. Resolved.

dataland-qa-service/src/main/kotlin/org/dataland/datalandqaservice/repositories/DataPointQaReportRepository.kt
florian-hoeft: Okay. In that case we can leave the comments as they are.

dataland-qa-service/src/main/kotlin/org/dataland/datalandqaservice/repositories/DataPointQaReportRepository.kt
florian-hoeft: Again, in that case we can leave the comment as is. Resolved.

dataland-qa-service/src/main/kotlin/org/dataland/datalandqaservice/repositories/DataPointQaReportRepository.kt
florian-hoeft: Thanks. Resolved.

dataland-qa-service/src/main/kotlin/org/dataland/datalandqaservice/services/DataPointQaReportManager.kt
ivan-gila: done

dataland-qa-service/src/main/kotlin/org/dataland/datalandqaservice/repositories/DataPointQaReportRepository.kt
florian-hoeft: Thanks. Resolved.

dataland-qa-service/src/test/kotlin/org/dataland/datalandqaservice/services/QaReviewTestFixtures.kt
florian-hoeft: Thanks. Resolved.

dataland-qa-service/src/test/kotlin/org/dataland/datalandqaservice/services/QaReviewTestFixtures.kt
florian-hoeft: Thanks. Resolved

dataland-qa-service/src/test/kotlin/org/dataland/datalandqaservice/org/dataland/datalandqaservice/services/DataPointQaReportManagerTest.kt
```diff
@@ -0,0 +1,58 @@
...
+import org.mockito.kotlin.verify
+import org.mockito.kotlin.whenever
+
+@TestInstance(TestInstance.Lifecycle.PER_CLASS)
+class DataPointQaReportManagerTest {
+    private val mockDataPointControllerApi: DataPointControllerApi = mock()
+    private val mockRepo: DataPointQaReportRepository = mock()
+    private val mockSecurityPolicy: QaReportSecurityPolicy = mock()
+
+    private val manager =
```
florian-hoeft: please rename manager to datapointQaReportManager

dataland-qa-service/src/test/kotlin/org/dataland/datalandqaservice/org/dataland/datalandqaservice/services/DataPointQaReportManagerTest.kt
```diff
@@ -0,0 +1,58 @@
...
+    fun `returns empty map for empty input`() {
+        val result = manager.countQaReportsForDataPointIdsBulk(emptySet())
+
+        assertEquals(emptyMap<String, Long>(), result)
+        verify(mockRepo, never()).countByDataPointIdInGrouped(any())
+    }
+
+    @Test
+    fun `splits into batches and aggregates results`() {
+        val manyIds: Set<String> = (0..50000).map { "id-$it" }.toSet()
```
florian-hoeft: Please use MAX_DATA_POINT_IDS_PER_BATCH from the actual DataPointQaReportManager here. Also, make it clear to whoever is reading the test is that manyIds is a set that is 50001 entries big and thus, it is above the limit MAX_DATA_POINT_IDS_PER_BATCH which causes the function countQaReportsForDataPointIdsBulk to batch.

dataland-qa-service/src/test/kotlin/org/dataland/datalandqaservice/org/dataland/datalandqaservice/services/DataPointQaReportManagerTest.kt
```diff
@@ -0,0 +1,58 @@
...
+
+        whenever(mockRepo.countByDataPointIdInGrouped(any())).doAnswer { invocation ->
```
florian-hoeft: Could you simplify this part of the test by stubbing two calls explicitly and verify both batching and aggregation. So:

whenever(mockRepo.countByDataPointIdInGrouped(any()))
.thenAnswer {
listOf(object : DataPointCount {
override fun getDataPointId(): String = "shared-id"
override fun getCount(): Long = 1L
})
}
.thenAnswer {
listOf(object : DataPointCount {
override fun getDataPointId(): String = "shared-id"
override fun getCount(): Long = 1L
})
}


This would make the test easier to understand.

florian-hoeft (commented):
2nd round review (additional comments for added test).

dataland-qa-service/src/test/kotlin/org/dataland/datalandqaservice/org/dataland/datalandqaservice/services/DataPointQaReportManagerTest.kt
ivan-gila: done

dataland-qa-service/src/test/kotlin/org/dataland/datalandqaservice/org/dataland/datalandqaservice/services/DataPointQaReportManagerTest.kt
ivan-gila: done

dataland-qa-service/src/test/kotlin/org/dataland/datalandqaservice/org/dataland/datalandqaservice/services/DataPointQaReportManagerTest.kt
ivan-gila: done

dataland-qa-service/src/test/kotlin/org/dataland/datalandqaservice/org/dataland/datalandqaservice/services/DataPointQaReportManagerTest.kt
florian-hoeft: Thanks. Resolved.

dataland-qa-service/src/test/kotlin/org/dataland/datalandqaservice/org/dataland/datalandqaservice/services/DataPointQaReportManagerTest.kt
florian-hoeft: Thanks. Resolved.

dataland-qa-service/src/test/kotlin/org/dataland/datalandqaservice/org/dataland/datalandqaservice/services/DataPointQaReportManagerTest.kt
florian-hoeft: Thanks. Resolved.

PR #1912, opened 2026-05-06
DALA-7300 WSL-based Development Process and Agentic AI Workflow Setup

braun-steven:
# Pull Request \<Title>
`<Description here>`


dataland-frontend/AGENTS.md
```diff
@@ -0,0 +1,44 @@
...
+# Commands
+
+- Install dependencies: `npm install`
+- Start local dev server: `npm run dev`
+- Format files: `npm run format`
+- Check formatting without changes: `npm run formatci`
+- Lint files: `npm run lint`
+- Lint in CI mode: `npm run lintci`
+- Type-check: `npm run typecheck`
+- Build production bundle: `npm run build`
```
StephanWezorke: This might not work. With shared elements and website there are two dependencies which are only respected by gradle. Even if the agent runs the build commands in the dependencies first, the files are only copied by gradle tasks.

dataland-frontend/AGENTS.md
```diff
@@ -0,0 +1,44 @@
...
+- When changing Cypress tests, support code, or fixture generation, also run the relevant Cypress compilation checks.
+- When changing user-facing behavior, run the most relevant Cypress component or E2E tests.
+- If the change depends on backend API contracts, regenerate clients with `./gradlew :dataland-frontend:generateClients` from the repo root.
+
+# Package-Specific Rules
+
+- Do not introduce new `@ts-nocheck` directives in modified files.
+- Prefer existing shared components, composables, stores, and framework configuration patterns before adding new ones.
+- Keep styling aligned with PrimeVue defaults unless there is a clear need for custom scoped styles.
+- This package depends on `dataland-sharedElements` and on generated OpenAPI clients. Changes in either may require rebuilding this package.
```
StephanWezorke: It also depends on the website, which is not mentioned here.

.gitignore
```diff
@@ -47,4 +47,13 @@ keystore.jks
...
-/dataland-loki/data
\ No newline at end of file
+/dataland-loki/data
+
+# Ignore the contents of .opencode (has node_modules etc)
+.opencode/*
+
+# Allow opencode configs, skills and agents
+!.opencode/opencode.jsonc
+!.opencode/tui.jsonc
```
StephanWezorke: Why is there no draft for those files in the PR?

manageLocalStack.sh
```diff
@@ -8,14 +8,14 @@ source "$project_root/localstack/cert_functions.sh"
 
 print_usage() {
   echo "Usage: $(basename "$0") [--start] [--stop] [--reset] [--local-frontend] [--dev-env] [--self-signed-certs] [--simple] [--container-backend]"
-  echo "  --start: Start the development stack"
+  echo "  --start: Start the development stack (uses container backend by default)"
```
StephanWezorke: Omit the parenthesis and just highlight the non-default behaviour for the flag that runs the backend in the terminal.

manageLocalStack.sh
```diff
@@ -8,14 +8,14 @@ source "$project_root/localstack/cert_functions.sh"
...
   echo "  --stop: Stop the development stack"
-  echo "  --reset: Reset and restart the development stack from scratch"
+  echo "  --reset: Reset and restart the development stack from scratch (uses container backend by default)"
```
StephanWezorke: Omit the parenthesis and just highlight the non-default behaviour for the flag that runs the backend in the terminal.

manageLocalStack.sh
```diff
@@ -8,14 +8,14 @@ source "$project_root/localstack/cert_functions.sh"
...
   echo "  --local-frontend: Run in local frontend mode (redirect traffic to localhost)"
   echo "  --dev-env: Load environments/.env.dev before starting/resetting"
   echo "  --self-signed-certs: Generate and use self-signed SSL certificates instead of retrieving them"
   echo "  --simple: Shortcut for --dev-env --self-signed-certs --container-backend"
-  echo "  --container-backend: Run backend in Docker container instead of via Gradle bootRun"
+  echo "  --container-backend: Explicitly run backend in Docker container (this is enabled by default for --reset and --start as well)"
```
StephanWezorke: Since this is already the default, we should instead have a command for not running the backend in the container. If we don't want that at all, remove this flag.

opencode.jsonc
```diff
@@ -0,0 +1,45 @@
+{
+  "$schema": "https://opencode.ai/config.json",
+
+  // Default shared model for this repository.
+  "model": "github-copilot/gpt-5.4",
```
StephanWezorke: Why aren't we using a Claude model by default?

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,41 @@
+{
```
StephanWezorke, line 1: Why do we need to sync "Docker Desktop" with WSL and why do we need to run "Docker Desktop" in order to bring up the devcontainer? Isn't it possible to run it just within the WSL?

StephanWezorke (commented):
There is a lot of duplication across this and other instruction files. Not sure if needed. Not sure if worth to remove.
We should also include "The migration should be tested using a subclass of "BaseFlywayMigrationTest"." in the Backend Guidelines (Zeile 86), aber ich kann gerade keine Kommentare mehr hinzufügen.

opencode.jsonc
braun-steven: Sorry we haven't talked about a sensible default yet. I picked `gpt-5.4` as default since it doesn't immediately crush our usage limits, according to the rates we are charged. E.g. gpt-5.5 costs 7.5x of gpt-5.4 and Claude Opus 4.7 costs 15x that of gpt-5.4:

image

We should discuss this though. I just found out that on the first of June, copilot is also changing the billing to a token-based schema: [GitHub Copilot is moving to usage-based billing](https://github.blog/news-insights/company-news/github-copilot-is-moving-to-usage-based-billing/)

StephanWezorke: Yes, but Claude Sonnet 4.5/4.6 are at the same level as GPT-5.4 and I am always told the models were better. With the new pricing model, things may change, though. With the token-based pricing, gpt-5.5 is twice as expensive as gpt-5.4 and it seems like Sonnet is only slightly more expensive than gpt-4.5 while Opus is slightly more expensive than gpt-5.5.
This is only derived from tables; depending on how many iterations the agents perform, either model may burn tokens faster.
https://docs.github.com/en/copilot/reference/copilot-billing/models-and-pricing

.gitignore
braun-steven: Good catch -- I moved them to the opencode default project location `$PROJECT/opencode.jsonc` -- for `tui.jsonc`: we don't have/need that as of now. Fixed in https://github.com/d-fine/Dataland/commit/5b3d6831c549489af0660592af90b7c7ae497c93

dataland-frontend/AGENTS.md
braun-steven: This means we should remove this particular build instruction or should we even mention that the agent should not use this because of <explanation>?

dataland-frontend/AGENTS.md
braun-steven: Does `Changes in either may require rebuilding this package.` also hold true for the website w.r.t. frontend? Then I'll just add it here.

.devcontainer/devcontainer.json
braun-steven: New setup is:

Windows -> No more Docker Desktop now
WSL -> hosts Docker Engine in Linux
Devcontainer -> has docker-in-docker i.e., its own Docker Engine host to encapsulate the sandbox inside of the container -- the alternative would be docker-outside-of-docker which uses the WSL docker host, but then the sandbox can escape with elevated priviliges using docker (e.g. running `docker run --privilidged ...`)

opencode.jsonc
braun-steven: As discussed in the KT session yesterday, we now default to Sonnt 4.6 as it has the same pricing as gpt-5.4 right now (https://github.com/d-fine/Dataland/commit/6c62c115f13a62c2c7f31295a5ca3faa409a6b23)

localstack/docker_functions.sh
```diff
@@ -53,19 +89,19 @@ start_docker_services() {
...
   fi
 }
 
 clear_docker_completely() {
-  echo "Clearing Docker..."
-  docker compose --profile development --profile developmentContainerFrontend --profile developmentContainerBackend down
-  docker compose --profile init down
-  docker compose down --remove-orphans
-  docker volume prune --force --all
+  log_step "Clearing Docker state"
```
StephanWezorke: What is the reason to sometimes use `log_step_done` and sometimes not?

localstack/docker_functions.sh
```diff
@@ -1,6 +1,40 @@
...
+run_step() {
+  local description="$1"
+  shift
+
+  log_step "$description"
+  run_quiet_command "$@"
+  log_step_done "$description"
+}
+
+docker_compose() {
```
StephanWezorke: I'd call this `run_docker_compose`. The underscore can be easily overlooked and future adjustments then may fall back to `docker compose`.

manageLocalStack.sh
```diff
@@ -81,25 +107,23 @@ check_backend_not_running() {
...
-  set -x
-  ./verifyEnvironmentVariables.sh
+  run_step "Verifying environment variables" ./verifyEnvironmentVariables.sh
   check_backend_not_running
   clear_docker_completely
-  ./gradlew clean
+  run_step "Cleaning Gradle outputs" ./gradlew clean
   assemble_all_projects
   rebuild_gradle_dockerfile
   source_github_env_log
```
StephanWezorke: What is the reason for using step logs only on some commands?

localstack/logging_functions.sh
```diff
@@ -0,0 +1,129 @@
+#!/usr/bin/env bash
```
StephanWezorke, line 1: It is not directly evident, what the difference between a status and a step line is. Put some comments in here that also explain which functions to use in which cases. What is the public ABI, what are only private helpers?

dataland-frontend/AGENTS.md
StephanWezorke: Yes, it does.

dataland-frontend/AGENTS.md
StephanWezorke: I think we should tell it to run gradle tasks for building.

braun-steven: Okay, which command would that be? There is no specific build task for the frontend `build.gradle.kts`.

localstack/docker_functions.sh
braun-steven: Good point, done in 374ac696a800c3dc6d184a0f78c0947d32e84638

dataland-frontend/AGENTS.md
StephanWezorke: The task `:dataland-frontend:npm_run_build` depends on the build tasks in `dataland-website`. The sharedElements stuff is directly wired into `package.json`; it may require a separate build in sharedElements, but that should be coverd by
```
This package depends on `dataland-sharedElements` and on generated OpenAPI clients. Changes in either may require rebuilding this package.
```

dataland-frontend/AGENTS.md
braun-steven: Fixed in 5419bf8bdb90878f30ecedf34b5b12a035a8b63d

localstack/docker_functions.sh
braun-steven: We now only have `log_step` and `log_success` which simply have different prefixes.

localstack/logging_functions.sh
braun-steven: This should not be relevant after the logging simplification anymore. there is only log step/info/success/warn/error now which handles colors and prefixes (and stderr piping in case of error).

manageLocalStack.sh
braun-steven: Good point actually, the other commands were logging inside but now that you mention it, it is much clearer to treat these as the top-level steps so we don't pollute the rest with unnecessary logging. Fixed in 09f66d146d14ff3658f8005b880fe54ee8b77584

localstack/docker_functions.sh
```diff
@@ -52,9 +56,8 @@ determine_compose_profiles() {
 }
 
 stop_and_cleanup_containers() {
-  log_step "Cleaning up existing containers"
-  docker_compose --profile development --profile developmentContainerFrontend --profile developmentContainerBackend down
-  docker_compose --profile development --profile developmentContainerFrontend --profile developmentContainerBackend pull --ignore-pull-failures --include-deps
+  run_docker_compose "${development_profiles[@]}" down
+  run_docker_compose "${development_profiles[@]}" pull --ignore-pull-failures
```
StephanWezorke, line 58: Why did you remove the `--include-deps` flag?

docker-compose.local.yml
```diff
@@ -0,0 +1,8 @@
+services:
```
StephanWezorke: Why do we need a separate file for that? Can't we use docker compose profiles?

localstack/docker_functions.sh
braun-steven: Sorry, the reasoning was only given in another commit message:

https://github.com/d-fine/Dataland/pull/1912/changes/8c0742ce8df255155d7e7bbe879765de76b64fc9

I double checked with Siegfried and the consensus was that the flag is unnecessary here and basically does nothing.

docker-compose.local.yml
braun-steven: No, compose profiles only activate/deactivate services but what we need here is an override for the loki service to use a local docker volume instead. The issue with the current setup is that `$LOKI_VOLUME` is also used in the `healthCheck.sh` script and if we set `$LOKI_VOLUME` to a docker volume, the health check script loses access to the local health-check-log (currently points to `$LOKI_VOLUME/...`) which is used by alloy. This is fixable if we separate the actual place of where loki writes data to and where the health-check is stored in but that makes it even more complex.

braun-steven: So in principle `$LOKI_VOLUME` could point to a docker volume declared in the `docker-compose.yml` file but since `$LOKI_VOLUME` is used in the health-check script and assumed to point to an actual directory (not a docker volume) we cannot point `$LOKI_VOLUME` to a volume.

dataland-frontend/AGENTS.md
braun-steven: Fixed in https://github.com/d-fine/Dataland/pull/1912/changes/5419bf8bdb90878f30ecedf34b5b12a035a8b63d

PR #1951, opened 2026-06-15
DALA-7409 Streamline Dev Setup

braun-steven:
# Pull Request 
`<Description here>`
- [ ] The PR has been peer-reviewed
- [ ] Make sure you understand the entire workflow from the user’s perspective, and test/review it accordingly. If anything is unclear, please reach out to the dev team.


.github/workflows/loadEnvironmentVariables/action.yaml
```diff
@@ -0,0 +1,30 @@
...
+            exit 1
+          fi
+
+          # Load variables into the current shell
+          source "$env_file"
+
+          # Parse keys from this file and write the expanded values to $GITHUB_ENV
+          while IFS= read -r env_key; do
+            echo "$env_key=${!env_key}" >> "$GITHUB_ENV"
+          done < <(sed -nE 's/^([A-Za-z_][A-Za-z0-9_]*)=.*/\1/p' "$env_file")
```
StephanWezorke, line 29: This won't work with indentation (i.e. in if-blocks) or multiple definitions per line separated by semicolons. Should we put comments in the env-files explaining which formats are allowed?

.github/workflows/CI.yaml
```diff
@@ -324,54 +307,8 @@ jobs:
...
-          KEYCLOAK_JUDGE_VALUE: "a+lk/Ddwa5wWL9T6BTZMhdNGEBiX6dzngWr5ic4Yt44="
-          KEYCLOAK_JUDGE_SALT: "z7iDkOaNQfeUQEV3IBpEFQ=="
-          KEYCLOAK_JUDGE_PASSWORD: "password"
-          KEYCLOAK_PREMIUM_USER_VALUE: "Dic1AEUG0nKnZ8me/3GMxgfyUmbwmJwDWNXkC7arLcij2BDB0xeOgk8ZpfZPKmFNikr9Is5I4+Nyk3MB9zs7mA=="
-          KEYCLOAK_PREMIUM_USER_SALT: "K9nzg086pCNjq8dxl84GZg=="
-          KEYCLOAK_PREMIUM_USER_PASSWORD: "password"
-          KC_BOOTSTRAP_ADMIN_USERNAME: "admin"
-          KC_BOOTSTRAP_ADMIN_PASSWORD: "password"
           KEYCLOAK_GOOGLE_ID: "irrelevant_in_ci"
           KEYCLOAK_GOOGLE_SECRET: "irrelevant_in_ci"
```
StephanWezorke: Shouldn't the CI-specific env-variables be part of `.env.ci`?

environments/.env.ci
```diff
@@ -0,0 +1,19 @@
+# CI-specific overrides for values that intentionally differ from local development.
+
+# Proxy configuration
+DISABLE_HSTS=false
```
StephanWezorke: Is this a newly introduced variable? I didn't find a previous definition. If yes, why do we introduce it now?

environments/.env.ci
```diff
@@ -0,0 +1,19 @@
...
+
+# RabbitMQ configuration
+RABBITMQ_PASS=dataland
+RABBITMQ_PASS_HASH=1s3+3kZqRTAgidTV6NE+adO3ja5NFNQZHa9MBwhJ0Kx7Fw6F
+
+# Backend URLs
+INTERNAL_BACKEND_URL=http://backend:8080/api
```
StephanWezorke: Is this a newly introduced variable? I didn't find a previous definition. If yes, why do we introduce it now?

localstack/cert_functions.sh
```diff
@@ -3,12 +3,7 @@
 retrieve_ssl_certificates() {
   mkdir -p ./local/certs
 
-  local scp_args=()
-  if [[ "${SILENT:-false}" == true ]]; then
-    scp_args+=(-q)
-  fi
```
StephanWezorke, line 9: You introduced `VERBOSE` so why didn't you keep this one with inverted logic using the new variable?

localstack/logging_functions.sh
```diff
@@ -63,3 +64,81 @@ log_warn() {
...
+  local exit_code=0
+  if [[ "$VERBOSE" == true ]]; then
+    local output_file
+    output_file=$(mktemp)
+
+    if "$@" >"$output_file" 2>&1; then
+      exit_code=0
+    else
+      exit_code=$?
+    fi
```
StephanWezorke, line 100: Wouldn't
```bash
exit_code=$("$@" > "$output_file" 2>&1)
```
be sufficient? Or even
```bash
"$@" | tee -a "$MANAGE_LOCAL_STACK_LOG_FILE" 2>&1
exit_code=${PIPESTATUS[0]}
```
and omit the `cat`-lines below?

localstack/logging_functions.sh
```diff
@@ -63,3 +64,81 @@ log_warn() {
...
+
+    cat "$output_file"
+    cat "$output_file" >> "$MANAGE_LOCAL_STACK_LOG_FILE"
```
StephanWezorke, line 103: `tee -a "$MANAGE_LOCAL_STACK_LOG_FILE" < "$output_file"`

localstack/logging_functions.sh
```diff
@@ -63,3 +64,81 @@ log_warn() {
...
+    rm -f "$output_file"
+  else
+    if "$@" >> "$MANAGE_LOCAL_STACK_LOG_FILE" 2>&1; then
+      exit_code=0
+    else
+      exit_code=$?
+    fi
```
StephanWezorke, line 110: See above.

.gitignore
```diff
@@ -49,6 +49,9 @@ tls.crt
 # local logging data
 /dataland-loki/data
 
+# logging of local stack
+**/log/manageLocalStack.log
```
StephanWezorke: Why not a fixed directory?

build.gradle.kts
```diff
@@ -92,6 +97,34 @@ val normalizeFeCoverageForSonar by tasks.registering {
...
+    }
+    tasks.withType<Exec>().configureEach {
+        environment(devEnvironmentVariables)
+    }
+    tasks.withType<NpmTask>().configureEach {
+        environment.putAll(devEnvironmentVariables)
+    }
+    tasks.withType<NodeTask>().configureEach {
+        environment.putAll(devEnvironmentVariables)
+    }
```
StephanWezorke: This is a bit ugly. ChatGPT suggested
```Kotlin
tasks.configureEach {
    when (this) {
        is Test,
        is JavaExec,
        is Exec -> environment(devEnvironmentVariables)

        is NpmTask,
        is NodeTask -> environment.putAll(devEnvironmentVariables)
    }
}
```
Is that an option? By the way: Why is this needed? Can't the env-variables be injected via docker-compose?

.github/workflows/dockerLogs/action.yaml
```diff
@@ -20,7 +20,7 @@ runs:
         docker logs "$docker_service" > ./dockerLogs/${{ inputs.subfolder }}/"$docker_service".log 2>&1
         done
     - name: Upload docker log files
-      if: always()
+      if: ${{ always() && env.ACT != 'true' }}
```
StephanWezorke: Why is the dependency on `env.ACT` introduced?

.github/workflows/logObjects/action.yaml
```diff
@@ -70,6 +70,7 @@ runs:
         echo '${{ inputs.vars }}' > logs/$GITHUB_JOB/vars_json.log
         find logs/$GITHUB_JOB -type f -size -4c -delete
     - name: Store env and secrets
+      if: ${{ env.ACT != 'true' }}
```
StephanWezorke: Why is the dependency on `env.ACT` introduced?

manageLocalStack.sh
```diff
@@ -1,24 +1,25 @@
...
-  echo "Multiple options can be combined in any order. Execution order is: stop, reset, start"
+  echo "Usage: $(basename "$0") [--start] [--stop] [--reset] [--retrieve-certs] [--no-container-backend] [--verbose]" >&3
+  echo "  --start: Start the development stack" >&3
+  echo "  --stop: Stop the development stack" >&3
+  echo "  --reset: Reset and restart the development stack from scratch" >&3
+  echo "  --retrieve-certs: Retrieve the SSL certificates instead of using self-signed ones" >&3
+  echo "  --no-container-backend: Run backend without containers" >&3
+  echo "  --verbose: Print subcommand output while also writing it to the local stack log" >&3
+  echo "" >&3
+  echo "Multiple options can be combined in any order. Execution order is: stop, reset, start" >&3
```
StephanWezorke: The onboarding wiki is not adjusted yet.
Why isn't `--retrieve-certs` the default?

.github/workflows/dockerLogs/action.yaml
braun-steven: Sorry, that was an artifact from my tests with the local GitHub Action runner `act` which I forgot to remove, thanks for catching! Removed.

.github/workflows/logObjects/action.yaml
braun-steven: Sorry, that was an artifact from my tests with the local GitHub Action runner act which I forgot to remove, thanks for catching! Removed.

environments/.env.ci
braun-steven: No it's not new, the variable is defined in `.env.dev`: 

https://github.com/d-fine/Dataland/blob/1913d92b11a7986a50a1252cc70d0b3560a4b82d/environments/.env.dev#L65

and used in 

https://github.com/d-fine/Dataland/blob/1913d92b11a7986a50a1252cc70d0b3560a4b82d/docker-compose.yml#L62

I guess we could think about removing the default value in docker-compose.

So before, in the CI, the `DISABLE_HSTS` env var was not defined and `docker compose` then defaulted to `false` which is what we want for the CI. The new behavior explicitly sets the env var in the `.env.ci` vars to `false`.

localstack/logging_functions.sh
braun-steven: I used `tee` in the first implementation for this which lead to the process getting stuck somehow (I was a bit out of my bash-fu here), gpt-5.4 suggested to circument the new `tee` process and use cat instead. See this commit: 8fb12c11e67c7f25d51c0d2540a632350e653f0d

localstack/logging_functions.sh
braun-steven: See answer in https://github.com/d-fine/Dataland/pull/1951/changes#r3434130109

localstack/logging_functions.sh
braun-steven: See answer in https://github.com/d-fine/Dataland/pull/1951/changes#r3434130109

.gitignore
braun-steven: Good point, fixed and changed to `/log`

build.gradle.kts
braun-steven: > Why is this needed? Can't the env-variables be injected via docker-compose?

This ensures that the gradle task are always aware of the env vars, independent of docker compose and resolves issues where env vars were not properly loaded in the environment in which the gradle task was executed, specifcally: Intellij running gradle tasks. Unfortunately there is no way to tell Intellij "always load env vars from this file for this project". There are workaround for `Run` configurations, but this does not resolve the issue for running gradle tasks in Intellij.

> Is that an option? 

No, Kotlin doesn't statically group `NpmTask` and `NodeTask` as one concrete type that has the `environment` property, therefore this fails with

```
> Configure project :
e: file:///srv/data/dataland/build.gradle.kts:132:16: Unresolved reference. None of the following candidates is applicable because of receiver type mismatch:
public inline fun <T : AbstractExecTask<*>> AbstractExecTask<TypeVariable(T)>.environment(vararg environmentVariables: Pair<String, Any?>): TypeVariable(T) defined in org.gradle.kotlin.dsl
public inline fun JavaExec.environment(vararg environmentVariables: Pair<String, Any?>): JavaExec defined in org.gradle.kotlin.dsl
public inline fun Test.environment(vararg environmentVariables: Pair<String, Any?>): Test defined in org.gradle.kotlin.dsl
public inline fun ProcessForkOptions.environment(vararg environmentVariables: Pair<String, Any?>): ProcessForkOptions defined in org.gradle.kotlin.dsl
e: file:///srv/data/dataland/build.gradle.kts:136:16: Function invocation 'environment(...)' expected
e: file:///srv/data/dataland/build.gradle.kts:136:16: Unresolved reference. None of the following candidates is applicable because of receiver type mismatch:
public inline fun <T : AbstractExecTask<*>> AbstractExecTask<TypeVariable(T)>.environment(vararg environmentVariables: Pair<String, Any?>): TypeVariable(T) defined in org.gradle.kotlin.dsl
public inline fun JavaExec.environment(vararg environmentVariables: Pair<String, Any?>): JavaExec defined in org.gradle.kotlin.dsl
public inline fun Test.environment(vararg environmentVariables: Pair<String, Any?>): Test defined in org.gradle.kotlin.dsl
public inline fun ProcessForkOptions.environment(vararg environmentVariables: Pair<String, Any?>): ProcessForkOptions defined in org.gradle.kotlin.dsl

FAILURE: Build failed with an exception.
```

manageLocalStack.sh
braun-steven: > Why isn't --retrieve-certs the default?

Reverted to retrieve by default and add option `--self-signed-certs` back.

> The onboarding wiki is not adjusted yet.

Which part? I've updated all `./manageLocalStack.sh` calls / examples. There is still the deprecated non-WSL part in which this is mentioned but which is supposed to be removed anyway soon, no?

localstack/cert_functions.sh
braun-steven: `retrieve_ssl_certificates` and `setup_certificates` is now called using `run_step` which already redirects the output correctly depending on the `VERBOSE` mode, therefore it became unnecessary to use the additional check here.

.github/workflows/CI.yaml
braun-steven: Mh, if I understand correctly, it is not used in the CI, so maybe we should even remove it completly? I wouldn't put it in `.env.ci` if it's not used in the CI.

environments/.env.ci
braun-steven: No not new, but I simplified it now after looking at it again. It used to default to the following in `.env.dev`

https://github.com/d-fine/Dataland/blob/1913d92b11a7986a50a1252cc70d0b3560a4b82d/environments/.env.dev#L73-L74

which was for the non-containerized local stack. `./manageLocalStack.sh` then overrode it via

https://github.com/d-fine/Dataland/blob/1913d92b11a7986a50a1252cc70d0b3560a4b82d/manageLocalStack.sh#L63-L69 depending on the presence of `--no-container-backend`.

We now default to a `http://backend:8080/api` in `.env.dev` and just override it in `./manageLocalStack.sh` if `--no-container-backed` is set. No need for an additional override in `.env.ci` then.

.github/workflows/loadEnvironmentVariables/action.yaml
braun-steven: > Should we put comments in the env-files explaining which formats are allowed?

I think this makes sense, or do we expect / want them to key more complicated than `KEY=VALUE` (i.e. actual bash script with if-blocks etc)?

StephanWezorke: I don't really expect it, but I think having those comments telling what you're not allowed to do is worthwhile.

.github/workflows/CI.yaml
StephanWezorke: True, but with the current setup, the CI would inherit the values from `.env.dev` which may also not what we want? I put the comments on those two, but there are more variables defined in `CI.yaml`. Please double check.

environments/.env.ci
StephanWezorke: Then remove it from `docker-compose.yml`.

build.gradle.kts
StephanWezorke: And what about
```Kotlin
tasks.configureEach {
    when (this) {
        is Test,
        is JavaExec,
        is Exec -> environment(devEnvironmentVariables)
        is NpmTask -> environment.putAll(devEnvironmentVariables)
        is NodeTask -> environment.putAll(devEnvironmentVariables)
    }
}
```

.github/workflows/logObjects/action.yaml
StephanWezorke: Thanks, but now you introduced `if: always()` which isn't present on main.

environments/.env.ci
```diff
@@ -7,13 +7,13 @@ DISABLE_HSTS=false
...
 # Data Exporter
 DATA_EXPORTER_OUTPUT_DIRECTORY=/tmp
 
 # Logging configuration
 # E2E tests assert against early backend startup logs after long test runs.
 LOGGING_OPTIONS_MAX_SIZE=1G
 LOGGING_OPTIONS_MAX_FILE=1
+
+# Preapproval
+AUTO_PREAPPROVAL_QA_ACCEPTED_DATAPOINTS=true
```
StephanWezorke: Why is this introduced here? Was it specified before?

.github/workflows/loadEnvironmentVariables/action.yaml
braun-steven: Done.

build.gradle.kts
braun-steven: That works, we can even remove JavaExec since it inherits from Exec anyway.

~Done.~

braun-steven: Ah correction, `Test` and `Exec` cannot be smart-cast. We still need:


```kotlin
    tasks.configureEach {
        when (this) {
            is Test -> environment(devEnvironmentVariables)
            is Exec -> environment(devEnvironmentVariables)
            is NpmTask -> environment.putAll(devEnvironmentVariables)
            is NodeTask -> environment.putAll(devEnvironmentVariables)
        }
    }
```

Done.

.github/workflows/CI.yaml
braun-steven: > True, but with the current setup, the CI would inherit the values from .env.dev which may also not what we want?

`KEYCLOAK_GOOGLE_ID/SECRET` is not part of the `.env.dev` file -- or what do you mean here?

>  I put the comments on those two, but there are more variables defined in CI.yaml. Please double check.

I've now extracted additional env vars, that I missed earlier and moved them from `CI.yaml` to `.env.ci` (including `KEYCLOAK_GOOGLE_ID/SECRET`):

6cbdec946bf6d1b12cf70e967106c147ea054c9e

Done?

environments/.env.ci
braun-steven: Sorry, I meant to write you on teams. This fix was discussed with Siegfried/Florin in the "make CI green again" story and fixes a currently failing test since the var was missing from the CI.

.github/workflows/CI.yaml
```diff
@@ -307,20 +307,15 @@ jobs:
       - name: Execute E2E Tests
         run: ./testing/e2e_test_launcher.sh
         env:
-          KEYCLOAK_GOOGLE_ID: "irrelevant_in_ci"
-          KEYCLOAK_GOOGLE_SECRET: "irrelevant_in_ci"
           PROXY_IMAGE: ghcr.io/d-fine/dataland/dataland_inbound_proxy:${{ env.DATALAND_INBOUND_PROXY_VERSION }}
           CYPRESS_EXECUTION_ENVIRONMENT: "ci"
```
StephanWezorke: Do we want to have this in `.env.ci`, too? Can't we move every variable defined here that is not specified in terms of GitHub Secrets or matrix variables?

.github/workflows/CI.yaml
```diff
@@ -307,20 +307,15 @@ jobs:
...
           CYPRESS_TEST_GROUP: ${{ matrix.testGroup }}
           CYPRESS_IGNORE_EXTERNAL_STORAGE: "false"
```
StephanWezorke: Do we want to have this in `.env.ci`, too?

.github/workflows/CI.yaml
```diff
@@ -307,20 +307,15 @@ jobs:
...
           TEST_EXECUTOR: ${{ matrix.testExecutor }}
-          COMPOSE_DOCKER_CLI_BUILD: 1
-          DOCKER_BUILDKIT: 1
           MAILJET_API_ID: ${{ secrets.MAILJET_API_ID }}
           MAILJET_API_SECRET: ${{ secrets.MAILJET_API_SECRET }}
           INTERNAL_EURODATCLIENT_URL: "http://dummy-eurodat-client:8080"
```
StephanWezorke: Do we want to have this in `.env.ci`, too?

.github/workflows/CI.yaml
```diff
@@ -307,20 +307,15 @@ jobs:
...
           BACKEND_URL: ${{ secrets.BACKEND_URL }}
-          IGNORE_EXTERNAL_STORAGE_ERROR: "false"
           JAVA_TOOL_OPTIONS: "-Xms128m -Xmx384m"
```
StephanWezorke: Do we want to have this in `.env.ci`, too?

environments/.env.ci
StephanWezorke: On a second thought, the variable is replaced only for a single service in `docker-compose.yml` so making a global override in `.env.ci` has a different meaning. I propose to keep it in `docker-compose.yml` and remove it here.

.github/workflows/CI.yaml
braun-steven: No, this is specific to the e2e tests. We don't want to apply these JVM settings to all jobs.

.github/workflows/CI.yaml
braun-steven: Done.

.github/workflows/CI.yaml
braun-steven: Done.

.github/workflows/CI.yaml
braun-steven: Done.

.github/workflows/CI.yaml
braun-steven: Moved to .env.ci. Done.

environments/.env.ci
braun-steven: Done.

PR #2011, opened 2026-07-23
DALA-7159: DatasetReviewOverview Page: Use existing format of original datapoint column for other columns as well

florian-hoeft:
# Pull Request \<Title>
`<Description here>`
- [x] Make sure you understand the entire workflow from the user’s perspective, and test/review it accordingly. If anything is unclear, please reach out to the dev team.
- [x] At least 3 consecutive CI runs are successfully executed to ensure that there are no flaky tests.
- [x] Documentation is updated as required. If unsure whether or what to update, ask a fellow developer.
- [x] IF there are changes done to the framework data models or to a database entity class, the following steps were completed in order
- [x] ELSE, the new version is deployed to one of the dev servers using this branch
- [x] Confirm that the latest version (use the /gitinfo endpoint) of the feature branch is deployed to one of the dev servers (no longer enforced by GitHub)
- [x] Ensure that the release notes are written in a way that is understandable to a non-technical audience.


dataland-framework-toolbox/src/main/kotlin/org/dataland/frameworktoolbox/frameworks/lksg/custom/LksgSimpleCustomComponentBase.kt
```diff
@@ -43,6 +44,25 @@ open class LksgSimpleCustomComponentBase(
                     ),
                 ),
             ),
+            valueGetterByDataPoint =
```
julianhoelzdfine, line 47: I do not see any changes to the corresponding lksg ViewConfig.ts file. Is this correct?

dataland-framework-toolbox/src/main/kotlin/org/dataland/frameworktoolbox/frameworks/eutaxonomynonfinancials/custom/EuTaxonomyAssuranceComponent.kt
```diff
@@ -117,7 +141,7 @@ class EuTaxonomyAssuranceComponent(
                     setOf(
                         TypeScriptImport(
                             "formatAssuranceProviderForDataTable",
-                            "@/components/resources/dataTable/conversion/EutaxonomyAssuranceValueGetterFactory",
+                            formatterImportPath,
```
julianhoelzdfine, line 144: Shouldn't there also be an implementation of the valueGetterByData function or the AssuranceProvider data point?

julianhoelzdfine, line 144: valueGetterByDataPoint =
            FrameworkDisplayValueByDataPointLambda(
                "formatAssuranceProviderForDataTable(parseDataPoint(dataPoint) as AssuranceDataPoint)",
                setOf(
                    TypeScriptImport(
                        "formatAssuranceProviderForDataTable",
                        formatterImportPath,
                    ),
                    TypeScriptImport(
                        "parseDataPoint",
                        "@/components/resources/dataTable/conversion/DataPoints",
                    ),
                    TypeScriptImport(
                        "AssuranceDataPoint",
                        "@clients/backend",
                    ),
                ),
            ),

dataland-framework-toolbox/src/main/kotlin/org/dataland/frameworktoolbox/intermediate/components/YesNoComponent.kt
```diff
@@ -40,6 +41,27 @@ class YesNoComponent(
                 ),
                 label, getTypescriptFieldAccessor(),
             ),
+            valueGetterByDataPoint =
+                documentSupport.getFrameworkDisplayValueByDataPointLambda(
+                    FrameworkDisplayValueByDataPointLambda(
+                        "formatYesNoValueForDatatable(extractDatapointValue(dataPoint) as YesNoNa)",
```
julianhoelzdfine: Shouldn't this be cast as YesNo? Or is there a reason why it is cast as YesNoNa?

julianhoelzdfine (approved):
Review Backend

dataland-frontend/src/components/resources/datasetReview/DatasetReviewComparisonTable.vue
```diff
@@ -147,7 +147,13 @@
...
-                      <span class="main-text-color">
+                      <MultiLayerDataTableCell
+                        v-if="row.qaReviewDisplay.get(qaReporter.reporterUserId)"
+                        :content="row.qaReviewDisplay.get(qaReporter.reporterUserId) as AvailableMLDTDisplayObjectTypes"
+                        :meta-info="dataMetaInformation as DataMetaInformation"
+                        :inReviewMode="true"
+                      />
+                      <span v-else class="main-text-color">
                         {{ getQaDisplayText(row, qaReporter.reporterUserId) }}
                       </span>
```
DominikTschimmel-d-fine: The `v-if` branching duplicates the fallback decision  in both the template and the script. I would suggest to handle the logic entirely in the script functions in a single place. This would also reduce complexity. See comments below for my suggestions.

Here you could simply use 

```
 <MultiLayerDataTableCell
                        :content="row.qaReviewDisplay.get(qaReporter.reporterUserId) as AvailableMLDTDisplayObjectTypes"
                        :meta-info="dataMetaInformation as DataMetaInformation"
                        :inReviewMode="true"
                      />
```

dataland-frontend/src/components/resources/datasetReview/DatasetReviewComparisonTable.vue
```diff
@@ -166,7 +172,13 @@
...
-                      <span v-if="getCustomDisplayValue(row.dataPointTypeId) != null">
+                      <MultiLayerDataTableCell
+                        v-if="row.customDisplay"
+                        :content="row.customDisplay"
+                        :meta-info="dataMetaInformation as DataMetaInformation"
+                        :inReviewMode="true"
+                      />
+                      <span v-else-if="getCustomDisplayValue(row.dataPointTypeId) != null">
                         {{ getCustomDisplayValue(row.dataPointTypeId) }}
                       </span>
```
DominikTschimmel-d-fine: Similar to the [comment above](https://github.com/d-fine/Dataland/pull/2011/changes#r3710484564)

dataland-frontend/src/components/resources/datasetReview/DatasetReviewComparisonTable.vue
```diff
@@ -367,6 +384,122 @@ function buildRowsFromConfig(config: MLDTConfig<FrameworkData>, data: FrameworkD
...
+function getQaReviewFallbackDisplayForReport(qaReport: DataPointQaReport | undefined) {
+  const simpleText = getQaVerdictSimpleText(qaReport?.verdict);
+  let display: AvailableMLDTDisplayObjectTypes | undefined;
+  if (simpleText != null) {
+    display = formatStringForDatatable(simpleText);
+  } else {
+    display = formatStringForDatatable(getCorrectedDisplayFromQaReport(qaReport));
+  }
+  return display;
+}
```
DominikTschimmel-d-fine: Without the `v-if` condition in the template and with the fallback returning an `AvailableMLDTDisplayObjectTypes` object, this could be shortened significantly:

```
function getQaReviewDisplayForReport(
  qaReport: DataPointQaReport | undefined,
  valueGetterByDataPoint?: (dataPoint: string) => AvailableMLDTDisplayObjectTypes
): AvailableMLDTDisplayObjectTypes {
  switch (qaReport?.verdict) {
    case QaReportDataPointVerdict.QaAccepted:
      return formatStringForDatatable('QA Accepted');
    case QaReportDataPointVerdict.QaInconclusive:
      return formatStringForDatatable('QA Inconclusive');
    case QaReportDataPointVerdict.QaNotAttempted:
      return formatStringForDatatable('QA Not Attempted');
    default:
      if (valueGetterByDataPoint && qaReport?.correctedData != null) {
        return valueGetterByDataPoint(qaReport.correctedData);
      }
      return formatStringForDatatable(getCorrectedDisplayFromQaReport(qaReport) ?? '–');
  }
}
```

Note: the `??` operator essentially checks whether the result is `null`/`undefined` and returns `'–'` in that case.

dataland-frontend/src/components/resources/datasetReview/DatasetReviewComparisonTable.vue
```diff
@@ -367,6 +384,122 @@ function buildRowsFromConfig(config: MLDTConfig<FrameworkData>, data: FrameworkD
...
+      const qaReport = getQaReportFor(cell.dataPointTypeId, qaReporter.reporterUserId);
+      const display = getQaReviewDisplayForReport(qaReport, cell.valueGetterByDataPoint);
+      if (display != null) {
+        qaReviewMap.set(qaReporter.reporterUserId, display);
+      }
+    }
+  }
+  return qaReviewMap;
+
+}
```
DominikTschimmel-d-fine: Without the `v-if` structure and the fallback logic in [the comment below](https://github.com/d-fine/Dataland/pull/2011/changes#r3711204674), this could be reduced to a single branch:

```
function getQaReviewMap(cell: MLDTCellConfig<FrameworkData>): Map<string, AvailableMLDTDisplayObjectTypes> {
  const qaReviewMap = new Map<string, AvailableMLDTDisplayObjectTypes>();
  if (!cell.dataPointTypeId) return qaReviewMap;

  if (!cell.valueGetterByDataPoint) {
    console.log(
      `The data point with label ${cell.label} does not seem to have a valueGetterByDataPoint() function implemented. ` +
        `Falling back to using a simple string to display value.`
    );
  }

  for (const qaReporter of props.datasetReview.qaReporters) {
    const qaReport = getQaReportFor(cell.dataPointTypeId, qaReporter.reporterUserId);
    qaReviewMap.set(qaReporter.reporterUserId, getQaReviewDisplayForReport(qaReport, cell.valueGetterByDataPoint));
  }
  return qaReviewMap;
}
```

dataland-frontend/src/components/resources/datasetReview/DatasetReviewComparisonTable.vue
```diff
@@ -323,6 +336,8 @@ export type CellRow = {
   label: string;
   dataPointTypeId?: string;
   originalDisplay: AvailableMLDTDisplayObjectTypes;
+  qaReviewDisplay: Map<string, AvailableMLDTDisplayObjectTypes>;
+  customDisplay?: AvailableMLDTDisplayObjectTypes;
```
DominikTschimmel-d-fine: Without the `v-if` structure, the optional type can become a required type (i.e. without `?`).

dataland-frontend/src/components/resources/datasetReview/DatasetReviewComparisonTable.vue
```diff
@@ -367,6 +384,127 @@ function buildRowsFromConfig(config: MLDTConfig<FrameworkData>, data: FrameworkD
...
+  const reviewInfo = getReviewInfo(cell.dataPointTypeId);
+  if (reviewInfo?.customValue == null) {
+    return undefined;
+  }
+  if (!cell.valueGetterByDataPoint) {
+    // Fallback in case valueGetterByDataPoint is not implemented for this datapoint.
+    return formatStringForDatatable(getCustomDisplayValue(cell.dataPointTypeId));
+  }
+  return cell.valueGetterByDataPoint(reviewInfo.customValue);
+}
```
DominikTschimmel-d-fine, line 461: This could be shortened slightly and adapted to the version without `v-if` (fixed return type now):

```
function getCustomDisplay(cell: MLDTCellConfig<FrameworkData>): AvailableMLDTDisplayObjectTypes {
  const reviewInfo = getReviewInfo(cell.dataPointTypeId);
  const customValue = reviewInfo?.customValue;
  if (customValue == null) {
    return formatStringForDatatable('');
  }
  if (cell.valueGetterByDataPoint) {
    return cell.valueGetterByDataPoint(customValue);
  }
  return formatStringForDatatable(getCustomDisplayValue(cell.dataPointTypeId) ?? '');
}
```

dataland-framework-toolbox/src/main/kotlin/org/dataland/frameworktoolbox/frameworks/lksg/custom/LksgSimpleCustomComponentBase.kt
StephanWezorke: Lksg is not an assembled framework so there are no datapoints in Lksg and the whole `valueGetterByDataPoint`-logic is not supported.

dataland-framework-toolbox/src/main/kotlin/org/dataland/frameworktoolbox/frameworks/eutaxonomynonfinancials/custom/EuTaxonomyAssuranceComponent.kt
StephanWezorke: The ticket states, that this component can not be easily adapted to the new logic, because the display requires information about the surrounding data tree. The ticket lists two options, one of which is to fall back to the default raw text rendering.
Maybe the decision was made to use this simple fall back mechanism?

dataland-framework-toolbox/src/main/kotlin/org/dataland/frameworktoolbox/intermediate/components/YesNoComponent.kt
StephanWezorke: It seems like there is no `YesNo` component in the frontend, but `YesNoNa` exists. I am not sure why, let's discuss in the group.

dataland-framework-toolbox/src/main/kotlin/org/dataland/frameworktoolbox/frameworks/lksg/custom/LksgSimpleCustomComponentBase.kt
julianhoelzdfine: This doesn't make sense, as the valueGetterByDataPoint function doesn't recieve its inputs from the lksg dataset but from the dataPointJudgement object. Whether lksg is an assembled framework or not should not matter

dataland-frontend/src/components/resources/datasetReview/DatasetReviewComparisonTable.vue
```diff
@@ -574,19 +718,8 @@ function isRowEmpty(cellRow: CellRow): boolean {
...
-    case QaReportDataPointVerdict.QaInconclusive:
-      return 'QA Inconclusive';
-    case QaReportDataPointVerdict.QaNotAttempted:
-      return 'QA Not Attempted';
-    default:
-      return getCorrectedDisplayFromQaReport(report) ?? '–';
-  }
+  const report = cellRow.dataPointTypeId ? getQaReportFor(cellRow.dataPointTypeId, reporterUserId) : undefined;
+  return getQaVerdictSimpleText(report?.verdict) ?? getCorrectedDisplayFromQaReport(report) ?? '–';
 }
```
DominikTschimmel-d-fine: Without the `v-if` structure, this should no longer be needed.

dataland-framework-toolbox/src/main/kotlin/org/dataland/frameworktoolbox/frameworks/eutaxonomynonfinancials/custom/EuTaxonomyAssuranceComponent.kt
julianhoelzdfine: It this datapoint is dependent on surrounding data, wouldn't be wise to let this be a calculated field? Or maybe I am misunderstood.

dataland-framework-toolbox/src/main/kotlin/org/dataland/frameworktoolbox/intermediate/components/YesNoComponent.kt
julianhoelzdfine: Thanks.

dataland-frontend/src/components/resources/datasetReview/DatasetReviewComparisonTable.vue
```diff
@@ -421,13 +559,13 @@ function getRowDataTest(row: KpiRow): string | undefined {
...
- * @param {CellRow} row - The table cell row describing the data point.
+ * @param {string} dataPointTypeId - The data point type id that describes the data type of the data point.
  * @param {string} reporterUserId - The userId of the user who uploaded the QA report.
  * @returns {DataPointQaReport | undefined} The matching QA report summary or undefined when not found.
  */
-function getQaReportFor(row: CellRow, reporterUserId: string): DataPointQaReport | undefined {
-  if (!row.dataPointTypeId) return undefined;
-  const datapointEntry = props.datasetReview.dataPoints[row.dataPointTypeId];
+function getQaReportFor(dataPointTypeId: string, reporterUserId: string): DataPointQaReport | undefined {
+  if (!dataPointTypeId) return undefined;
```
DominikTschimmel-d-fine: Allow optional arguments  (otherwise the `if (!dataPointTypeId)` seems pointless) and handle the `undefined` return here (see [this comment below](https://github.com/d-fine/Dataland/pull/2011/changes#r3714048232) and [this comment below](https://github.com/d-fine/Dataland/pull/2011/changes#r3714053477))

```
function getQaReportFor(dataPointTypeId?: string, reporterUserId?: string): DataPointQaReport | undefined {
  if (!dataPointTypeId || !reporterUserId) return undefined;
```

dataland-frontend/src/components/resources/datasetReview/DatasetReviewComparisonTable.vue
```diff
@@ -515,7 +653,10 @@ function isCellEmpty(cellRow: CellRow, source: AcceptedDataPointSource, reporter
     }
 
     case AcceptedDataPointSource.Qa: {
-      const report = reporterUserId == null ? undefined : getQaReportFor(cellRow, reporterUserId);
+      const report =
+        reporterUserId == null || !cellRow.dataPointTypeId
+          ? undefined
+          : getQaReportFor(cellRow.dataPointTypeId, reporterUserId);
```
DominikTschimmel-d-fine: With [the above change](https://github.com/d-fine/Dataland/pull/2011/changes#r3714041633), this can simply be

```
const report = getQaReportFor(cellRow.dataPointTypeId, reporterUserId);
```

dataland-frontend/src/components/resources/datasetReview/DatasetReviewComparisonTable.vue
```diff
@@ -543,7 +684,10 @@ function shouldShowRejectedIcon(cellRow: CellRow, source: AcceptedDataPointSourc
   if (isCellEmpty(cellRow, source, reporterUserId)) return false;
   if (isAcceptedSource(cellRow, source, reporterUserId)) return false;
   if (source === AcceptedDataPointSource.Qa) {
-    const report = reporterUserId == null ? undefined : getQaReportFor(cellRow, reporterUserId);
+    const report =
+      reporterUserId == null || !cellRow.dataPointTypeId
+        ? undefined
+        : getQaReportFor(cellRow.dataPointTypeId, reporterUserId);
```
DominikTschimmel-d-fine: With [the above change](https://github.com/d-fine/Dataland/pull/2011/changes#r3714041633), this can simply be

```
const report = getQaReportFor(cellRow.dataPointTypeId, reporterUserId);
```

DominikTschimmel-d-fine (commented):
Note: These changes may require updates to the KDocs, which were not addressed in the individual comments.

dataland-framework-toolbox/src/main/kotlin/org/dataland/frameworktoolbox/intermediate/components/YesNoComponent.kt
StephanWezorke: This is just about the formatting of the value. Since YesNo is a subset of YesNoNa, the YesNoNa-formatter can be used for YesNo and no additional file is needed.

dataland-framework-toolbox/src/main/kotlin/org/dataland/frameworktoolbox/intermediate/components/ViewConfigCells.kt
```diff
@@ -0,0 +1,118 @@
+package org.dataland.frameworktoolbox.intermediate.components
```
StephanWezorke, line 1: This file is a bit hard to read. Here are a few suggestions for improvements:

- The first two functions have identical parameters (up to default values) and long parameter lists. Use a DTO, you could also put only parts of the parameters in the DTO and for instance leave the `sectionConfigBuilder` as is.
- `dataPointValueExpression` seems to always fall back to `extractDatapointValue` (document support) or `parseDataPoint` (no document support), so do we really need this as a dynamic parameter?
- Creation of `FrameworkDisplayValueLambda` and `FrameworkDisplayValueByDataPointLambda` is identical in the first two functions. Could we combine the functions and use a boolean flag for document support?
```Kotlin
val valueGetter = FrameworkDisplayValueLambda(...).let {
    documentSupport.getFrameworkDisplayValueLambda(it, ...) if documentSuppoert else it
}
val rawValueGetterByDataPoint = ... // similar
```
It may be worthwhile to have a helper function for that.
- What is the purpose of `fieldAccessor`? It seems to be unused downstream. Can we remove it?
- The function names are not giving much insights. From what they do, I'd expect the second function to be named similar to the first one, but with "no document support" or the like. The third one is essentially the first one with different parameters, so why not just use the same name?

dataland-frontend/src/frameworks/sfdr/ViewConfig.ts
```diff
@@ -1381,7 +1381,7 @@ export const sfdrViewConfiguration: MLDTConfig<SfdrData> = [
           },
           {
             type: 'cell',
-            label: 'Water Consumption',
+            label: 'Water Withdrawal',
```
StephanWezorke, line 1384: You said the changes in this file are intended because otherwise the toolbox failed. But why did it fail? What is the reason we need this change?

dataland-framework-toolbox/src/test/kotlin/org/dataland/frameworktoolbox/intermediate/components/ViewConfigTestFixtures.kt
```diff
@@ -0,0 +1,38 @@
...
+ */
+fun assertGettersDifferOnlyInValueExpression(
+    cell: CellConfigBuilder,
+    datasetValueExpression: String,
+    dataPointValueExpression: String,
+) {
+    val placeholder = "<VALUE>"
+    assertEquals(
+        cell.valueGetter.lambdaBody.replace(datasetValueExpression, placeholder),
+        cell.valueGetterByDataPoint!!.lambdaBody.replace(dataPointValueExpression, placeholder),
```
StephanWezorke: Get into the habit of avoiding `!!`. In this case you can just access the fields using `?.` and the assertion will fail if the field is null.

dataland-framework-toolbox/src/test/kotlin/org/dataland/frameworktoolbox/intermediate/components/ViewConfigTestFixtures.kt
```diff
@@ -0,0 +1,38 @@
+package org.dataland.frameworktoolbox.intermediate.components
```
StephanWezorke: This is more like a utility file than a file generating fixtures. Also: Can you use a static class to make the functions namespaced?

dataland-framework-toolbox/src/test/kotlin/org/dataland/frameworktoolbox/frameworks/eutaxonomynonfinancials/custom/EuTaxonomyNonAlignedActivitiesComponentTest.kt
```diff
@@ -0,0 +1,57 @@
...
+            "$FORMATTER(dataset.revenue?.$FIELD,\"My Field\", \"revenue\")",
+            cell.valueGetter.lambdaBody,
+        )
+        assertEquals(
+            "$FORMATTER(parseDataPoint(dataPoint) as Parameters<typeof $FORMATTER>[0], \"My Field\", \"revenue\")",
+            cell.valueGetterByDataPoint!!.lambdaBody,
+        )
+
+        assertTrue(cell.valueGetterByDataPoint!!.usesDataPoint)
+        assertFalse(cell.valueGetterByDataPoint!!.usesDataset)
```
StephanWezorke: Get into the habit of avoiding `!!`. In this case you can just access the fields using `?.` and the assertion will fail if the field is null.

dataland-framework-toolbox/src/test/kotlin/org/dataland/frameworktoolbox/frameworks/eutaxonomynonfinancials/custom/EuTaxonomyNonAlignedActivitiesComponentTest.kt
```diff
@@ -0,0 +1,57 @@
...
+
+        val section = newSection()
+        component.generateDefaultViewConfig(section)
+
+        // Characterisation test, not intended behaviour: the kpi type is derived via
+        // `getTypescriptFieldAccessor().split(".")[1].dropLast(1)`, which is only correct while the component is
+        // nested in a ComponentGroup (where `dropLast(1)` strips the trailing "?" of the null-safe accessor).
+        // Without a group parent the accessor is `dataset.<field>` and the last character of the field name is eaten.
+        val cell = onlyCell(section)
+        listOf(cell.valueGetter.lambdaBody, cell.valueGetterByDataPoint!!.lambdaBody).forEach { body ->
```
StephanWezorke: `!!`

dataland-framework-toolbox/src/test/kotlin/org/dataland/frameworktoolbox/frameworks/eutaxonomynonfinancials/custom/EuTaxonomyNonAlignedActivitiesComponentTest.kt
```diff
@@ -0,0 +1,57 @@
...
+        val component =
+            EuTaxonomyNonAlignedActivitiesComponent(FIELD, DemoComponentGroupApiImpl()).also { it.label = "My Field" }
+
+        val section = newSection()
+        component.generateDefaultViewConfig(section)
+
+        // Characterisation test, not intended behaviour: the kpi type is derived via
+        // `getTypescriptFieldAccessor().split(".")[1].dropLast(1)`, which is only correct while the component is
+        // nested in a ComponentGroup (where `dropLast(1)` strips the trailing "?" of the null-safe accessor).
+        // Without a group parent the accessor is `dataset.<field>` and the last character of the field name is eaten.
```
StephanWezorke: Why are we testing unintended behaviour?

dataland-frontend/src/frameworks/sfdr/ViewConfig.ts
julianhoelzdfine: While I was working on the ticket my framework-toolbox broke and produced the label 'Water Consumption' instead of 'Water Withdrawal'. After fixing it it produced the right label again. 

A recent ticket required the renaming of some kpi names in the sfdr framework, which would have constituted api breaking changes. Therefore an intermediate fix was to introduce a search and replace in the framework toolbox until customers where informed about the change. My framework toolbox must have failed before the search and replace could run.

dataland-framework-toolbox/src/test/kotlin/org/dataland/frameworktoolbox/intermediate/components/ViewConfigTestFixtures.kt
julianhoelzdfine: Thanks. I didn't know this. I changed this.

dataland-framework-toolbox/src/main/kotlin/org/dataland/frameworktoolbox/intermediate/components/ViewConfigCells.kt
julianhoelzdfine: > The first two functions have identical parameters (up to default values) and long parameter lists. Use a DTO, you could also put only parts of the parameters in the DTO and for instance leave the sectionConfigBuilder as is.

Eine DTO wurde eingeführt.

> dataPointValueExpression seems to always fall back to extractDatapointValue (document support) or parseDataPoint (no document support), so do we really need this as a dynamic parameter?

The expression differs in the cast type that is handed over. Now the dataPointValueExpression is hardcoded in the functions but still need the specification of the cast type. This make the function call site of add...ValueCell slightly more beautiful

> Creation of FrameworkDisplayValueLambda and FrameworkDisplayValueByDataPointLambda is identical in the first two functions. Could we combine the functions and use a boolean flag for document support?

Ich denke nicht, dass dadurch Lesbarkeit gewonnen werden kann. Der Body der Funktion ist dann einfach mit einem if statements unlesbarer geworden.

> What is the purpose of fieldAccessor? It seems to be unused downstream. Can we remove it?

It is only unused in the getFrameworkDisplayValueByDataPointLambda and used in the getFrameworkDisplayValueLambda path. I have cleaned it up from the former path. This was a known oversight, but for some reason forgotten to remove.

> The function names are not giving much insights. From what they do, I'd expect the second function to be named similar to the first one, but with "no document support" or the like. The third one is essentially the first one with different parameters, so why not just use the same name?

Guter Punkt. Ich habe die zweite Funktion nun addNonDocumentSupportedValueCell genannt.

dataland-framework-toolbox/src/test/kotlin/org/dataland/frameworktoolbox/intermediate/components/ViewConfigTestFixtures.kt
julianhoelzdfine: The file has been renamed and the functions are now namespaced

dataland-framework-toolbox/src/test/kotlin/org/dataland/frameworktoolbox/frameworks/eutaxonomynonfinancials/custom/EuTaxonomyNonAlignedActivitiesComponentTest.kt
julianhoelzdfine: Good Question. In the framework toolbox this is not really necessary. I have removed the test.

dataland-framework-toolbox/src/test/kotlin/org/dataland/frameworktoolbox/frameworks/eutaxonomynonfinancials/custom/EuTaxonomyNonAlignedActivitiesComponentTest.kt
julianhoelzdfine: It was replaced. However, the the assert function does not allow null to be handed over. The null case needs to be handled manually.

dataland-frontend/src/components/resources/datasetReview/DatasetReviewComparisonTable.vue
julianhoelzdfine: I have removed the if-statement here.

Althought this had as a consequence that I can not return

if (!cell.dataPointTypeId) return emptyMap

in the definition of `getQaReviewMap`, as then the get statement of `row.qaReviewDisplay.get(...)` will return undefined if `dataPointTypeId ` is null and the `qaReporters `list is non-empty.

Currently, I have decided to return a map with `qaReporter.reporterUserId` as keys and `formatStringForDatatable('')` as values. So the get always return a valid `AvailableMLDTDisplayObjectTypes`

dataland-frontend/src/components/resources/datasetReview/DatasetReviewComparisonTable.vue
julianhoelzdfine: Has been changed.

dataland-framework-toolbox/src/test/kotlin/org/dataland/frameworktoolbox/frameworks/lksg/custom/LksgSimpleCustomComponentBaseTest.kt
```diff
@@ -0,0 +1,67 @@
...
+        component("My Field").generateDefaultViewConfig(section)
+
+        val cell = onlyCell(section)
+        assertEquals(
+            "$FORMATTER(dataset.subcontractingCompanies, \"My Field\")",
+            cell.valueGetter.lambdaBody,
+        )
+        assertEquals(
+            "$FORMATTER(parseDataPoint(dataPoint) as Parameters<typeof $FORMATTER>[0], \"My Field\")",
+            cell.valueGetterByDataPoint!!.lambdaBody,
```
StephanWezorke: !!

dataland-framework-toolbox/src/test/kotlin/org/dataland/frameworktoolbox/frameworks/lksg/custom/LksgSimpleCustomComponentBaseTest.kt
```diff
@@ -0,0 +1,67 @@
...
+        assertTrue(cell.valueGetter.imports.contains(TypeScriptImport(FORMATTER, FORMATTER_MODULE)))
+        assertTrue(cell.valueGetterByDataPoint!!.imports.contains(TypeScriptImport(FORMATTER, FORMATTER_MODULE)))
+        assertTrue(
+            cell.valueGetterByDataPoint!!.imports.contains(
+                TypeScriptImport("parseDataPoint", "@/components/resources/dataTable/conversion/DataPoints"),
+            ),
+        )
+
+        assertTrue(cell.valueGetterByDataPoint!!.usesDataPoint)
+        assertFalse(cell.valueGetterByDataPoint!!.usesDataset)
```
StephanWezorke: !!

dataland-framework-toolbox/src/test/kotlin/org/dataland/frameworktoolbox/frameworks/lksg/custom/LksgSimpleCustomComponentBaseTest.kt
```diff
@@ -0,0 +1,67 @@
...
+            "$FORMATTER(dataset.subcontractingCompanies, \"My Field\")",
+            cell.valueGetter.lambdaBody,
+        )
+        assertEquals(
+            "$FORMATTER(parseDataPoint(dataPoint) as Parameters<typeof $FORMATTER>[0], \"My Field\")",
+            cell.valueGetterByDataPoint!!.lambdaBody,
+        )
+
+        // This component hand-writes both getters instead of routing through ViewConfigCells, so a copy-paste of the
+        // wrong formatter name is the realistic regression here.
```
StephanWezorke: What is the point of this comment?

dataland-framework-toolbox/src/test/kotlin/org/dataland/frameworktoolbox/intermediate/components/ComponentViewConfigTest.kt
```diff
@@ -0,0 +1,150 @@
+package org.dataland.frameworktoolbox.intermediate.components
+
+import org.dataland.frameworktoolbox.intermediate.components.ViewConfigTestUtils.assertGettersDifferOnlyInValueExpression
+import org.dataland.frameworktoolbox.intermediate.components.ViewConfigTestUtils.newSection
+import org.dataland.frameworktoolbox.intermediate.components.ViewConfigTestUtils.onlyCell
```
StephanWezorke: Import only `ViewConfigTestUtils` and use `ViewConfigTestUtils.function(...)` in the code.

dataland-frontend/src/components/resources/datasetReview/DatasetReviewComparisonTable.vue
julianhoelzdfine: This does not have the same problem as above and was easier to fix. I have removed the if statement here.

dataland-frontend/src/components/resources/datasetReview/DatasetReviewComparisonTable.vue
julianhoelzdfine: This was a good idea and is now implemented.

dataland-frontend/src/components/resources/datasetReview/DatasetReviewComparisonTable.vue
julianhoelzdfine: This is also a good idea and is now implemented.

dataland-frontend/src/components/resources/datasetReview/DatasetReviewComparisonTable.vue
julianhoelzdfine: This has also been changed.

dataland-frontend/src/components/resources/datasetReview/DatasetReviewComparisonTable.vue
julianhoelzdfine: Has been removed

dataland-framework-toolbox/src/test/kotlin/org/dataland/frameworktoolbox/intermediate/components/ComponentViewConfigTest.kt
```diff
@@ -0,0 +1,150 @@
...
+        listOf(cell.valueGetter.lambdaBody, cell.valueGetterByDataPoint!!.lambdaBody).forEach { body ->
+            assertContains(body, mappingPrelude)
+        }
+        assertGettersDifferOnlyInValueExpression(cell, DATASET_ACCESSOR, dataPointExpression)
+
+        // ViewConfig.ts.ftl picks the lambda parameter via FrameworkLambda.usesDataset / usesDataPoint, which are
+        // case-sensitive `lambdaBody.contains(...)` checks. This body embeds `getDatasetAsMap` and
+        // `DropdownDatasetIdentifier`, which only escape `usesDataset` because of their capital "D". Introducing a
+        // lower-case `dataset` here (e.g. by renaming the helper or adding a local) would make the template emit
+        // `(dataset: XData)` for a getter that is actually called with a data-point string.
```
StephanWezorke: Let's discuss, I want to understand that in more detail.

dataland-frontend/src/components/resources/datasetReview/DatasetReviewComparisonTable.vue
julianhoelzdfine: Has been implemented

dataland-framework-toolbox/src/test/kotlin/org/dataland/frameworktoolbox/intermediate/components/ComponentViewConfigTest.kt
```diff
@@ -0,0 +1,150 @@
...
+        component<FreeTextComponent>().generateDefaultViewConfig(section)
+
+        val cell = onlyCell(section)
+        assertEquals("formatFreeTextForDatatable($DATASET_ACCESSOR)", cell.valueGetter.lambdaBody)
+        assertEquals(
+            "formatFreeTextForDatatable((extractDatapointValue(dataPoint) as string))",
+            cell.valueGetterByDataPoint!!.lambdaBody,
+        )
+        assertFalse(cell.valueGetter.imports.contains(EXTRACT_DATAPOINT_VALUE_IMPORT))
+        assertTrue(cell.valueGetterByDataPoint!!.imports.contains(EXTRACT_DATAPOINT_VALUE_IMPORT))
```
StephanWezorke: I wrote this elsewhere but maybe a bit more reasoning: Either you want to test if the field is not null, then you should introduce a dedicated check for that and not rely on the implicitly failing `!!` without making the problem clear. Or you don't need the test, because you certainly know the field can not be `null`. Then just use the null-safe field accessor so the test intent becomes evident.

There are more occurrences of `!!` in this file.

dataland-frontend/src/components/resources/datasetReview/DatasetReviewComparisonTable.vue
julianhoelzdfine: Has been changed

dataland-frontend/src/components/resources/datasetReview/DatasetReviewComparisonTable.vue
julianhoelzdfine: Has been changed

dataland-framework-toolbox/src/test/kotlin/org/dataland/frameworktoolbox/intermediate/components/ViewConfigCellsTest.kt
```diff
@@ -0,0 +1,182 @@
+package org.dataland.frameworktoolbox.intermediate.components
+
+import org.dataland.frameworktoolbox.intermediate.components.ViewConfigTestUtils.newSection
+import org.dataland.frameworktoolbox.intermediate.components.ViewConfigTestUtils.onlyCell
```
StephanWezorke: Import only `ViewConfigTestUtils` and use `ViewConfigTestUtils.function(...)` in the code.

dataland-framework-toolbox/src/test/kotlin/org/dataland/frameworktoolbox/intermediate/components/ViewConfigCellsTest.kt
```diff
@@ -0,0 +1,182 @@
...
+        assertEquals(
+            "formatStringForDatatable((extractDatapointValue(dataPoint) as string))",
+            cell.valueGetterByDataPoint!!.lambdaBody,
+        )
+        assertEquals(cell.valueGetter.imports + EXTRACT_DATAPOINT_VALUE_IMPORT, cell.valueGetterByDataPoint!!.imports)
+        assertEquals(cell.valueGetter.imports + cell.valueGetterByDataPoint!!.imports, cell.imports)
+    }
+
+    @Test
+    fun `addDocumentSupportedValueCell - check if both getters are build only the datapoint getter references dataPoint`() {
```
StephanWezorke: Check if what?

dataland-framework-toolbox/src/test/kotlin/org/dataland/frameworktoolbox/intermediate/components/ViewConfigCellsTest.kt
```diff
@@ -0,0 +1,182 @@
...
+
+        val nonDocumentSupportedSection = newSection()
+        stringComponent(NoDocumentSupport).addNonDocumentSupportedValueCell(
+            nonDocumentSupportedSection,
+            ValueGetterSpec(formatterImports = emptySet(), dataPointCastType = "MyType") { it },
+        )
+        val nonDocumentSupportedGetter = onlyCell(nonDocumentSupportedSection).valueGetterByDataPoint!!
+
+        // The reader function is derived from the called function, so it can no longer be mismatched with the
+        // import that same function adds.
```
StephanWezorke: Ehm.. what?

dataland-framework-toolbox/src/test/kotlin/org/dataland/frameworktoolbox/intermediate/components/ViewConfigCellsTest.kt
```diff
@@ -0,0 +1,182 @@
...
+        assertEquals("(extractDatapointValue(dataPoint) as MyType)", documentSupportedGetter.lambdaBody)
+        assertTrue(documentSupportedGetter.imports.contains(EXTRACT_DATAPOINT_VALUE_IMPORT))
+        assertFalse(documentSupportedGetter.imports.contains(PARSE_DATAPOINT_IMPORT))
+
+        assertEquals("(parseDataPoint(dataPoint) as MyType)", nonDocumentSupportedGetter.lambdaBody)
+        assertTrue(nonDocumentSupportedGetter.imports.contains(PARSE_DATAPOINT_IMPORT))
+        assertFalse(nonDocumentSupportedGetter.imports.contains(EXTRACT_DATAPOINT_VALUE_IMPORT))
```
StephanWezorke: What is really tested here? The import checks are identical, so the only difference is about `parseDataPoint` vs `extractDataPoint`. Isn't that already covered by the other functions?

dataland-framework-toolbox/src/test/kotlin/org/dataland/frameworktoolbox/frameworks/eutaxonomynonfinancials/custom/EuTaxonomyNonAlignedActivitiesComponentTest.kt
```diff
@@ -29,29 +28,11 @@ class EuTaxonomyNonAlignedActivitiesComponentTest {
...
         assertEquals(
             "$FORMATTER(parseDataPoint(dataPoint) as Parameters<typeof $FORMATTER>[0], \"My Field\", \"revenue\")",
-            cell.valueGetterByDataPoint!!.lambdaBody,
+            cell.valueGetterByDataPoint?.lambdaBody,
         )
 
-        assertTrue(cell.valueGetterByDataPoint!!.usesDataPoint)
-        assertFalse(cell.valueGetterByDataPoint!!.usesDataset)
+        assertTrue(cell.valueGetterByDataPoint?.usesDataPoint ?: false)
+        assertFalse(cell.valueGetterByDataPoint?.usesDataset ?: true)
```
StephanWezorke: We're getting into style discussions, but personally, I think one of 
```Kotlin
assertEquals(true, cell.valueGetterByDataPoint?.usesDataPoint)
assertTrue(cell.valueGetterByDataPoint?.usesDataPoint == true)
```
is much easier to read. I'd suggest using the first one.

dataland-framework-toolbox/src/test/kotlin/org/dataland/frameworktoolbox/frameworks/eutaxonomynonfinancials/custom/EuTaxonomyNonAlignedActivitiesComponentTest.kt
julianhoelzdfine: Irrelevant, since test was deleted.

dataland-framework-toolbox/src/main/kotlin/org/dataland/frameworktoolbox/frameworks/eutaxonomynonfinancials/custom/EuTaxonomyAlignedActivitiesComponent.kt
```diff
@@ -28,5 +28,6 @@ class EuTaxonomyAlignedActivitiesComponent(
         EuTaxonomyActivitiesGenerationConfig(
             fixtureGeneratorMethodName = "generateAlignedActivity",
             specificationType = "EuTaxonomyAlignedActivitiesComponent",
+            uploadComponentNameOverride = "AlignedActivitiesFormField",
```
StephanWezorke, line 31: Why?

dataland-framework-toolbox/src/main/kotlin/org/dataland/frameworktoolbox/intermediate/components/ViewConfigCells.kt
```diff
@@ -12,84 +12,96 @@ private val EXTRACT_DATAPOINT_VALUE_IMPORT =
...
- * value-getter, defaults to [getTypescriptFieldAccessor]
- * @param buildBody builds the lambda body from a value expression (either [datasetValueExpression] or
- * [dataPointValueExpression])
+ * @param valueGetterSpec describes how both value-getters are built; see [ValueGetterSpec]. The dataset-based
+ * value-getter reads the raw [getTypescriptFieldAccessor], i.e. the whole data point rather than its `.value`,
+ * because the formatter consumes the whole data point.
  */
-@Suppress("LongParameterList")
-fun ComponentBase.addParsedDataPointValueCell(
+fun ComponentBase.addNonDocumentSupportedValueCell(
```
StephanWezorke, line 88: Regarding the combination of the functions, I thought about something like (untested)
```Kotlin
private fun constructPlainValueGetters(
    valueGetterSpec: ValueGetterSpec,
    hasDocumentSupport: Boolean,
    datasetValueExpression: String,
): Pair<FrameworkDisplayValueLambda, FrameworkDisplayValueByDataPointLambda> {
    val dataPointValueExpression =
        "(${if (hasDocumentSupport) "extractDatapointValue" else "parseDataPoint"}(dataPoint) as ${valueGetterSpec.dataPointCastType})"

    val valueGetter = FrameworkDisplayValueLambda(valueGetterSpec.buildBody(datasetValueExpression), valueGetterSpec.formatterImports)
    val valueGetterByDataPoint =
        FrameworkDisplayValueByDataPointLambda(
            valueGetterSpec.buildBody(dataPointValueExpression),
            valueGetterSpec.formatterImports + PARSE_DATAPOINT_IMPORT + valueGetterSpec.additionalDataPointImports,
        )

    return Pair(valueGetter, valueGetterByDataPoint)
}

fun ComponentBase.addValueCell(
    sectionConfigBuilder: SectionConfigBuilder,
    valueGetterSpec: ValueGetterSpec,
    hasDocumentSupport: Boolean,
) = constructPlainValueGetters(
    valueGetterSpec,
    hasDocumentSupport,
    getTypescriptFieldAccessor(hasDocumentSupport),
).run {
    if (hasDocumentSupport) {
        Pair(
            documentSupport.getFrameworkDisplayValueLambda(first, label, getTypescriptFieldAccessor()),
            documentSupport.getFrameworkDisplayValueByDataPointLambda(second, label),
        )
    } else {
        this
    }
}.let {
    sectionConfigBuilder.addStandardCellWithValueGetterFactory(this, it.first, it.second)
}
```
It reduces the duplication, but is not necessarily better to read. Decide on your own which one you want to use.

dataland-frontend/src/components/resources/datasetReview/DatasetReviewComparisonTable.vue
```diff
@@ -392,54 +387,28 @@ function buildRowsFromConfig(config: MLDTConfig<FrameworkData>, data: FrameworkD
...
  */
 function getQaReviewMap(cell: MLDTCellConfig<FrameworkData>): Map<string, AvailableMLDTDisplayObjectTypes> {
-  const qaReviewMap = new Map<string, AvailableMLDTDisplayObjectTypes>();
   if (!cell.dataPointTypeId) {
-    return qaReviewMap;
+    // create a Map that displays an empty string for each qaReporter. If this is missing we get a NullPointerException.
+    return new Map(
+      props.datasetReview.qaReporters.map((qaReporter) => [
+        qaReporter.reporterUserId,
+        formatStringForDatatable(NO_DATA_POINT_ID_FOUND_LABEL),
```
DominikTschimmel-d-fine, line 395: Until now, the behavior was to display `'-'` (i.e. `NO-DATA-POINT-VALUE_FOUND_LABEL`) even for cells without `dataPointTypeId`. Is this an intentional change, here?

dataland-frontend/src/components/resources/datasetReview/DatasetReviewComparisonTable.vue
```diff
@@ -490,19 +446,18 @@ function getQaReviewFallbackDisplayForReport(
...
+function getCustomDisplay(cell: MLDTCellConfig<FrameworkData>): AvailableMLDTDisplayObjectTypes {
+  if (!cell.dataPointTypeId) {
+    return formatStringForDatatable(NO_DATA_POINT_ID_FOUND_LABEL);
   }
-  const reviewInfo = getReviewInfo(cell.dataPointTypeId);
-  if (reviewInfo?.customValue == null) {
-    return undefined;
+  const customValue = getReviewInfo(cell.dataPointTypeId)?.customValue;
+  if (!customValue) {
+    return formatStringForDatatable(NO_DATA_POINT_VALUE_FOUND_LABEL);
```
DominikTschimmel-d-fine, line 455: Until now, the custom value cell remained empty (i.e. `''` or now `NO_DATA_POINT_ID-FOUND_LABEL`). With this change, the custom value column would be filled with dashes at the beginning of the review. Is that the intended behavior?

dataland-frontend/src/components/resources/datasetReview/DatasetReviewComparisonTable.vue
```diff
@@ -490,19 +446,18 @@ function getQaReviewFallbackDisplayForReport(
...
   }
-  if (!cell.valueGetterByDataPoint) {
-    // Fallback in case valueGetterByDataPoint is not implemented for this datapoint.
-    return formatStringForDatatable(getCustomDisplayValue(cell.dataPointTypeId));
+  if (cell.valueGetterByDataPoint) {
+    return cell.valueGetterByDataPoint(customValue);
   }
-  return cell.valueGetterByDataPoint(reviewInfo.customValue);
+  return formatStringForDatatable(getCustomDisplayValue(cell.dataPointTypeId) ?? NO_DATA_POINT_VALUE_FOUND_LABEL);
```
DominikTschimmel-d-fine, line 460: See above

dataland-frontend/src/components/resources/datasetReview/DatasetReviewComparisonTable.vue
DominikTschimmel-d-fine: `undefined` is no longer a return option.

dataland-frontend/src/components/resources/datasetReview/DatasetReviewComparisonTable.vue
```diff
@@ -490,19 +446,18 @@ function getQaReviewFallbackDisplayForReport(
  * @param {MLDTCellConfig<FrameworkData>} cell - The cell configuration to build the custom display for.
  * @returns {AvailableMLDTDisplayObjectTypes | undefined} The display value, or undefined when unavailable.
```
DominikTschimmel-d-fine: `undefined` is no longer a return option.

dataland-frontend/src/components/resources/datasetReview/DatasetReviewComparisonTable.vue
julianhoelzdfine: Up until now this behavior would have thrown an error.

dataland-frontend/src/components/resources/datasetReview/DatasetReviewComparisonTable.vue
julianhoelzdfine: It was the other way arround. When no data point is found `''` is displayed.
Changing from `''` to `'-'` was intentional to make the behavior in the custom column analogous to the qa column

dataland-frontend/src/components/resources/datasetReview/DatasetReviewComparisonTable.vue
julianhoelzdfine: Dito

dataland-frontend/src/components/resources/datasetReview/DatasetReviewComparisonTable.vue
julianhoelzdfine: Was removed.

dataland-frontend/src/components/resources/datasetReview/DatasetReviewComparisonTable.vue
julianhoelzdfine: Was removed.

dataland-framework-toolbox/src/test/kotlin/org/dataland/frameworktoolbox/frameworks/lksg/custom/LksgSimpleCustomComponentBaseTest.kt
julianhoelzdfine: Removed

dataland-framework-toolbox/src/test/kotlin/org/dataland/frameworktoolbox/frameworks/lksg/custom/LksgSimpleCustomComponentBaseTest.kt
julianhoelzdfine: Removed

dataland-framework-toolbox/src/test/kotlin/org/dataland/frameworktoolbox/frameworks/lksg/custom/LksgSimpleCustomComponentBaseTest.kt
julianhoelzdfine: The comment has been removed

dataland-framework-toolbox/src/test/kotlin/org/dataland/frameworktoolbox/intermediate/components/ComponentViewConfigTest.kt
julianhoelzdfine: This was also changed in the LksgCustomComponentBaseTest file.

dataland-framework-toolbox/src/test/kotlin/org/dataland/frameworktoolbox/intermediate/components/ComponentViewConfigTest.kt
julianhoelzdfine: Comment has been removed.

dataland-framework-toolbox/src/test/kotlin/org/dataland/frameworktoolbox/intermediate/components/ComponentViewConfigTest.kt
julianhoelzdfine: Has been removed

dataland-framework-toolbox/src/test/kotlin/org/dataland/frameworktoolbox/frameworks/eutaxonomynonfinancials/custom/EuTaxonomyNonAlignedActivitiesComponentTest.kt
julianhoelzdfine: I went with `assertNotNull` beforehand and used the non-nullable object it returns.

dataland-framework-toolbox/src/test/kotlin/org/dataland/frameworktoolbox/intermediate/components/ViewConfigCellsTest.kt
julianhoelzdfine: Has been changed in all four test files

dataland-framework-toolbox/src/test/kotlin/org/dataland/frameworktoolbox/intermediate/components/ViewConfigCellsTest.kt
julianhoelzdfine: Comment removed.

dataland-framework-toolbox/src/main/kotlin/org/dataland/frameworktoolbox/intermediate/components/ViewConfigCells.kt
julianhoelzdfine: this might make it more compact, but I do not think that this improves readability. I would leave as is

dataland-framework-toolbox/src/main/kotlin/org/dataland/frameworktoolbox/frameworks/eutaxonomynonfinancials/custom/EuTaxonomyAlignedActivitiesComponent.kt
julianhoelzdfine: I don't know. This change has been done by Florian. Please ask him on monday.

dataland-framework-toolbox/src/test/kotlin/org/dataland/frameworktoolbox/intermediate/components/ViewConfigCellsTest.kt
julianhoelzdfine: I fixed the title of the test. It was missing a conjuction.

dataland-framework-toolbox/src/test/kotlin/org/dataland/frameworktoolbox/intermediate/components/ViewConfigCellsTest.kt
julianhoelzdfine: You are right. The test is mostly redundant and its test cases are covered by other tests. I have removed it.

dataland-frontend/src/frameworks/sfdr/ViewConfig.ts
florian-hoeft: I presume this is resolved because it says both "label: Water Withdrawal" on main and on this branch?

florian-hoeft: So I investigated this a bit further. I reran my framework toolbox for "sfdr" and the ViewConfig.ts was modified. 
The following labels were changed after my run: 
"OLD -> NEW "
1) Water Withdrawal -> Water Consumption 
2) Water Withdrawal Intensity -> Relative Water Usage 
3) Violation of UNGC principles and OECD Guidelines for Multinational Enterprises -> Violation Of Tax Rules And Regulation
4) Risk of Child Labour Incidents -> Reported Child Labour Incidents
...

I double checked the input file "sfdr.xlsx" and indeed, those labels were updated. 
Since we did not touch the input file here, my instinct would be to simply let the framework toolbox run and push whatever is produced? I do not know which api breaking change Julian is referring too.

dataland-framework-toolbox/src/main/kotlin/org/dataland/frameworktoolbox/frameworks/eutaxonomynonfinancials/custom/EuTaxonomyAlignedActivitiesComponent.kt
florian-hoeft: The framework toolbox previously did not run through for frameworks that inlcuded this component. This was an old bug that pre-dated our ticket I believe. The issue was simply that the type "EuTaxonomyActivitiesGenerationConfig" has the non-optional parameter "uploadComponentNameOverride" that was missing here. 

So to get the toolbox running, we added the parameter here.

dataland-framework-toolbox/src/main/kotlin/org/dataland/frameworktoolbox/intermediate/components/ViewConfigCells.kt
florian-hoeft: I would also leave them as they are

PR #2057, opened 2026-08-19
DALA-6377 Decommission edit data set functionality

JoshuaKoeck:
DALA-6377: Decommission edit data set functionality

## Things to do during Peer Review
Please check all boxes before the Pull Request is merged. In case you skip a box, describe in the PRs description (that means: here) why the check is skipped.
- [x] The GitHub Actions for the CI are green. This is enforced by GitHub. 
- [x] The PR has been peer-reviewed
  - [x] The code has been manually inspected by someone who did not implement the feature
  - [x] The changes have been inspected for possible breaking API changes (changed data models, endpoints, response codes, exception handling, ...). If there are any discuss them with the team.
  - [x] If this PR includes work on the frontend, at least one `@ts-nocheck` is removed. Additionally, there should not be any `@ts-nocheck` in files modified by this PR. If no `@ts-nocheck` are left: Celebrate :tada: :confetti_ball: type-safety and remove this entry. 
- [x] The PR actually implements what is described in the JIRA-Issue
- [x] Make sure you understand the entire workflow from the user’s perspective, and test/review it accordingly. If anything is unclear, please reach out to the dev team.
- [x] At least one test exists testing the new feature
  - [x] Make sure all newly added functionality (especially user facing) is tested
  - [x] Make sure that new test files are included in a test container and actually run in the CI
- [ ] At least 3 consecutive CI runs are successfully executed to ensure that there are no flaky tests.
- [x] Documentation is updated as required. If unsure whether or what to update, ask a fellow developer.
- [x] If there was a database entity class added, there must also be a migration script for creating the corresponding database if flyway is already used by the service
- [ ] IF there are changes done to the framework data models or to a database entity class, the following steps were completed in order
  - [ ] A fresh clone of dataland.com is generated (see Wiki page on "OTC" for details)
  - [ ] The feature branch is deployed to clone with `Reset non-user related Docker Volumes & Re-populate` turned off
  - [ ] It's verified that the CD run is green
  - [ ] The new feature is manually used/tested/observed on the clone server. Testing of the new feature should (also) be done by a second, independent reviewer from the dev team
  - [ ] The feature branch is deployed to one of the dev servers with `Reset non-user related Docker Volumes & Re-populate` turned on, and it's verified that the CD run is green
- [x] ELSE, the new version is deployed to one of the dev servers using this branch
  - [x] Run with setting `Reset non-user related Docker Volumes & Re-populate` turned on 
[...]

PR #2058, opened 2026-08-19
DALA 4405 Remove document meta data from framework data models

florian-hoeft:
# Pull Request \<Title>
`<Description here>`
- [ ] The PR has been peer-reviewed
- [ ] Make sure you understand the entire workflow from the user’s perspective, and test/review it accordingly. If anything is unclear, please reach out to the dev team.


dataland-backend/src/main/kotlin/org/dataland/datalandbackend/configurations/ApiClients.kt
```diff
@@ -38,12 +38,14 @@ class ApiClients(
...
-     * Creates an auto-authenticated version of the DocumentControllerApi of the document manager
+     * Creates an auto-authenticated version of the DocumentControllerApi of the document manager.
+     * Uses the patient (long-timeout) OkHttp client because batch metadata requests for large
+     * portfolios can take longer than the default timeout to complete.
      */
     @Bean
     fun getDocumentControllerApi(
-        @Qualifier("AuthenticatedOkHttpClient") authenticatedOkHttpClient: OkHttpClient,
-    ): DocumentControllerApi = DocumentControllerApi(documentManagerBaseUrl, authenticatedOkHttpClient)
+        @Qualifier("PatientAuthenticatedOkHttpClient") patientAuthenticatedOkHttpClient: OkHttpClient,
```
PauliKogler, line 47: Why exactly 10 minutes for the PatientAuthenticatedOkHttpClient timeout? That seems like a lot. Isn't there a middle ground, like 30-60s?
If the document-manager call fails slowly (e.g. it's stuck or the DB is having issues), does that mean the request just hangs for up to 10 minutes before it fails? That seems bad for a user-facing GET.
Also: does this timeout apply only to the batch call, or to all DocumentControllerApi calls (e.g. the single-document lookup too)?

dataland-backend/src/test/kotlin/org/dataland/datalandbackend/utils/DataDeliveryServiceUtilsTest.kt
```diff
@@ -0,0 +1,196 @@
...
+                """{"value":0.5,"dataSource":{"page":"1","fileName":null,"fileReference":"$enrichableFileReference"}}""",
+            )
+
+        val result = dataDeliveryServiceUtils.enhanceDataPoints(mapOf("id" to dataPoint), emptyMap())
+
+        assertFalse(
+            result.allStoredDataPoints
+                .getValue("id")
+                .dataPoint
+                .contains("\"fileName\":\"$documentName\""),
```
PauliKogler: This test's assertion is too weak: it only checks that fileName isn't equal to the specific documentName string used in another test, not that it's actually still null. Since the mock already returns name = null, this check would still pass even if the code wrongly wrote something like "" or a placeholder instead of keeping null. Better to directly assert that dataSource.fileName is null (e.g. by parsing the JSON), so the test really checks what its name says.

dataland-backend/src/main/kotlin/org/dataland/datalandbackend/configurations/ApiClients.kt
florian-hoeft: The timeout applies to all calls made through the DocumentControllerApi. 
I think setting it indivually for each api call is a bit overkill, so I have set it to 150s for now.

dataland-backend/src/test/kotlin/org/dataland/datalandbackend/utils/DataDeliveryServiceUtilsTest.kt
florian-hoeft: Correct. I edited the test. 
The JSON is now parsed, and we check if the fieldName field is still null after enrichment.

dataland-e2etests/src/test/kotlin/org/dataland/e2etests/tests/dataPoints/DataMigrationTest.kt
```diff
@@ -215,6 +206,12 @@ class DataMigrationTest {
                 dataId = dataMetaInfo.dataId,
                 body = linkedQaReportData.qaReport,
             )
+        if (testDataLocation.contains("WithManyNulls")) {
```
PauliKogler, line 213: This test silently expects a ServerException for the "many nulls" fixture, but there's no comment explaining why. Please add a comment explaining the reason.

PauliKogler:
The DataIntegrity.ts files (
dataland-frontend/tests/e2e/specs/lksg/DataIntegrity.ts
dataland-frontend/tests/e2e/specs/sfdr/DataIntegrity.ts
dataland-frontend/tests/e2e/specs/pcaf/DataIntegrity.ts
dataland-frontend/tests/e2e/specs/eu-taxonomy-financials/DataIntegrity.ts
dataland-frontend/tests/e2e/specs/eu-taxonomy-non-financials/DataIntegrity.ts
) were deleted on main in #2057 because the "edit dataset" feature was removed. But during this merge, they ended up as empty files instead of being deleted. This looks like a merge mistake . These files should probably just be removed instead of kept empty.

dataland-e2etests/src/test/kotlin/org/dataland/e2etests/tests/dataPoints/DataMigrationTest.kt
florian-hoeft: I added a comment why this was done. 
Here is a summary: With the ticket, we do not allow fileName and publicationDate to be set in the dataSource field of a datapoint when uploading anymore. 
During this migration test, fileName/publicationDate are stripped from datapoints before uploading. This can lead to entire "dataSource" fields becoming completely null. The backend then ignores those fields, which later triggers a comparison to fail. 

I would keep this as is for now, even though it is not 100% clean code.

